PR #3, opened 2021-02-09
[WAYANG-1][WAYANG-2] Parent artifact and modification code structure

berttty:
correction on the code structure and also generation of a parent module


wayang-api/pom.xml
```diff
@@ -15,94 +15,109 @@
...
-            </plugin>
-        </plugins>
-    </build>
 
     <dependencies>
         <dependency>
             <groupId>org.apache.wayang</groupId>
             <artifactId>wayang-core</artifactId>
-            <version>1.0.0-SNAPSHOT</version>
+            <version>${wayang.core.version}</version>
```
chrisdutz: Using expressions like this "<version>${wayang.core.version}</version>" tend to cause more problems than they solve ... I would strongly reccomend to stick to using plain-text versions. The Release-plugin updates them for you anyway.

chrisdutz (changes requested):

wayang-api/pom.xml
berttty: Ok, I will change it

berttty:
I did the changes on the variables

ro-pardo:
Test are running locally

PR #4, opened 2021-02-09
Changes Module API

berttty:
I solve the issue [WAYANG-7] and [WAYANG-11] to enable the development of Python API


ro-pardo:
The changes are correct, now we can continue with the API modules.

PR #9, opened 2021-02-23
Python api

ro-pardo:
Python Wayang first version, has the most essential components just running together very simple plans. It is still missing operators, UDFs and serialization methods.


berttty:
The protoc path need to be environment variable to easy change in the future :) and add some reference in the documentation. please take as todo :)

PR #11, opened 2021-03-23
Python api

ro-pardo:
Now supports juncture topologies, also reduce the number of sockets used by wayang by nesting python operator functions on the same pipeline.


berttty:
The pull request have the following issues.

-  have absolute path to your machine
- Change the print to logs records, you can find help [here](https://docs.python.org/3/howto/logging.html)

ro-pardo:
Corrections has been applied

berttty:
I cancel the auto build because they new got in a queue

berttty:
Everything looks ok :D

PR #22, opened 2021-09-03
[WAYANG-35] the library Trove4j was removed

berttty:
It's important to focus on the TODO added because they may affect the performance of some code


atroudi:
Looks good to merge.

PR #44, opened 2021-12-01
[CI] add skywalking-eyes check and fix some file license

CalvinKirs:


.licenserc.yaml
```diff
@@ -0,0 +1,40 @@
...
+    - .gitattributes
+    - '**/*.md'
+    - '**/*.json'
+    - '**/*.iml'
+    - '**/*.input'
+    - '**/*.lock'
+    - '**/*.ini'
+    - '**/.gitignore'
+    - '**/org/apache/wayang/commons/util/profiledb/**'
+
```
CalvinKirs, line 39: It is best to modify all files under this package to jar dependencies

CalvinKirs:
local check: https://github.com/CalvinKirs/incubator-wayang/runs/4378043946?check_suite_focus=true

Is GithubAction activated?

berttty:
Currently we don't have activated the githubaction, but we can do it :D

.licenserc.yaml
berttty: Because we are using Apache Rat to check the licenses, and the header in that file did not throw any exception because have the elements that Apache Rat Check, however, that does means that is a correct one, if you have a better one or a recomendation we can change it

CalvinKirs: My suggestion here is to modify it as a jar dependency, (if possible)
It is important to keep NOTICE as brief and simple as possible, as each addition places a burden on downstream consumers.

Here are some introductions, you can take a look:https://github.com/apache/skywalking-eyes

CalvinKirs:
HI, any feedback?

berttty:
I'm sorry for the delay. I had been busy with personal stuff; during the weekend, I will activate the GitHub actions and merge your pull :D. I'm so happy that you are helping us.

> HI, any feedback?

berttty (commented):
everything looks perfect :D

PR #53, opened 2022-01-14
Add maven wrapper to ease the development

CalvinKirs:
Add maven wrapper that can ease the development (they don't need to install maven as long as they have Java environment), and we can lock the maven version so that there is no version issues in the future.


berttty:
Is it possible to do it without the binary file ? Like just downloading from internet

CalvinKirs:
> Is it possible to do it without the binary file ? Like just downloading from internet

hi, this PR has nothing to do with this.

CalvinKirs:
> > Is it possible to do it without the binary file ? Like just downloading from internet
> 
> hi, this PR has nothing to do with this.

We can see similar functionality in other apache projects like camel.

berttty:
Ok :D

PR #54, opened 2022-01-15
[CI]Add UT check 

CalvinKirs:
We have used the skywalking-eyes check license head, so we can remove RAT.

### DIFF
--- a/.github/workflows/ut.yml
+++ b/.github/workflows/ut.yml
@@ -21,9 +21,6 @@ on:
     branches:
       - main
 
-env:
-  LOG_DIR: /tmp/wayang
-
 concurrency:
   group: unit-test-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
--- a/.github/workflows/ut.yml
+++ b/.github/workflows/ut.yml
@@ -1,0 +1,35 @@
+name: Unit Test
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+
+env:
+  LOG_DIR: /tmp/wayang
+
+concurrency:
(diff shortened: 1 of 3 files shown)


CalvinKirs:
I noticed that we have some check tools that still use the `RAT`, but are not affected by this PR change.

ro-pardo:
@CalvinKirs Hi, seems very well. But it is nice to have the RAT for an early checking of the licenses while developing.
Could you preserve it in addition to your work?

CalvinKirs:
> 

hi, currently we only use it to detect license headers in our Travis CI, but we have used `SkyWalking-Eyes` for detection (I will add check license dep later), its function is more powerful, and the community is very active. So I think we can remove the `RAT`. What do you think?

berttty:
I guess what @ro-pardo is point out, it is when you are working in your local machine, because the `SkyWalking-Eyes` is focusing after you did the push, and the `RAT` it will analyze when your are working in your local machine. 

Do am I wrong @ro-pardo  ??

ro-pardo:
@berttty That is exactly my point. Could it be inconvenient to only check the licenses after the push. Of course is great to add this new function, for sure has many other uses; but i think there is no reason to remove RAT by now.

CalvinKirs:
> @berttty That is exactly my point. Could it be inconvenient to only check the licenses after the push. Of course is great to add this new function, for sure has many other uses; but i think there is no reason to remove RAT by now.

OK, I've restored the RAT, thanks for the reminder.

berttty (commented):
Everything looks nice

ro-pardo (commented):
Implementation seems correct

PR #55, opened 2022-01-15
Delete maven-wrapper.jar

ro-pardo:
Not required, able to download jar binaries in execution time using https://github.com/apache/incubator-wayang/blob/main/mvnw#L202-L276.
This is to avoid future conflicts with binaries revisions.


berttty:
Nice to figure out 😄

PR #59, opened 2022-01-20
[CI]Auto Build And Install

CalvinKirs:


berttty:
I will double check the logs and I try to reproduce the error :D

CalvinKirs:
> I will double check the logs and I try to reproduce the error :D

I added the config file in UT, but, IMO,  it doesn't seem necessary

CalvinKirs:
Now we should only be left with the document update triggering.

PR #60, opened 2022-01-24
[CI] create Issues from the TODO in the code

berttty:
This incorporate the issue of repository [todo-to-issue-action](https://github.com/marketplace/actions/todo-to-issue) version [4.5.4](https://github.com/alstr/todo-to-issue-action/tree/v4.5.4) and the configuration to be execute as external repository


ro-pardo:
This will be merged because the Link appears to be in the Exceptions file

But for some reason is not taking it into account.

Nevertheless it works

PR #61, opened 2022-01-24
[CI] the action just in main branch

berttty:
The actions need to be executed when they occur in the 'main' branch


berttty:
the pull request does not correspond, because mix two of them

PR #62, opened 2022-01-24
[CI] execute the method just in main branch

berttty:
The actions need to be executed just when the element are push or some pull request is targeting the main branch but not in all the branch


ro-pardo:
This will be merged because the Link appears to be in the Exceptions file

But for some reason is not taking it into account.

Nevertheless it works

PR #74, opened 2022-02-11
[Github]Set protected branches and review restrictions

CalvinKirs:
Set protected branches and review restrictions
ignore todo files license head check


zhongjiajie (approved):
LGTM

PR #76, opened 2022-03-03
[BUG] delete paranamer-2.3 from dependecies

berttty:
delete paranamer-2.3 from dependecies to just work with paranamer 2.8, because exception at runtime in spark and other platforms


ro-pardo (approved):
Changes look correct

PR #77, opened 2022-03-04
upgrade plugins for Reproducible Builds

hboutemy:
tested 0.6.0 release, and given old versions of plugins, it's as expected not reproducible...

see https://maven.apache.org/guides/mini/guide-reproducible-builds.html#faq


berttty:
Thank you for catching it :D

berttty (approved):
The changes are reviewed

PR #78, opened 2022-03-07
Update README.md

regaleo605:
Add notes regarding issue with JAVA_HOME variable and setting up Wayang as well as specifying (current) supported Java version.


berttty (approved):
Thanks for catching it :D

PR #212, opened 2022-04-11
JVM-platform inside of Python-API

berttty:
Python API can run JVM platforms. However, these are basic elements that could help to it


JorgeQuiane (approved):
Everything looks ok. It is missing documentation but this will be addressed in future commits

PR #226, opened 2022-05-02
Correction in DistinctOperator for Apache Flink

### LINKED ISSUE #225: Operator Distinct One value
The distinct operator just shows one value independent of the value that could have

berttty:
The implementation of KeySelectorDistinct was having a issue in the implementation.

In addition to this the test for the Join and ReduceBy in the flink platform were disable because of possible issue in the implementation.

Close #225


zkaoudi (approved):
Everything is OK.

PR #235, opened 2022-06-13
Add Grep Benchmark

berttty:
Inclusion of the benchmark Grep in the benchmark module and other minors

### DIFF
--- a/wayang-benchmark/code/main/java/org/apache/wayang/apps/grep/Grep.java
+++ b/wayang-benchmark/code/main/java/org/apache/wayang/apps/grep/Grep.java
@@ -1,9 +1,26 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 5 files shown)


wayang-benchmark/code/main/java/org/apache/wayang/apps/grep/Grep.java
```diff
@@ -0,0 +1,150 @@
...
+
+  public static void wayangFlink(String input, String output){
+    wayangPlatform(input, output, Flink.basicPlugin());
+  }
+
+  public static void main(String... args) throws Exception {
+    int size = Integer.parseInt(args[0]);
+    String platform = args[1];
+
+    String input = args[2]+"/python/src/pywy/tests/resources/10e"+size+"MB.input";
```
JorgeQuiane, line 104: I understand the input path is now hard coded because it is not supposed to come from the user, yet I would suggest to have an input variable later on to allow for flexibility in the future.

JorgeQuiane (approved):
Just change the hard coded input later on (see my comment on File Grep.java)

PR #299, opened 2023-04-16
removed travis, added Apache Calcite support

2pk03:
minor changes, travis link failed since travis is decommissioned


zkaoudi (changes requested):
Apache Calcite is not a processing engine. So it should not be in the list of platforms that Wayang supports.

zkaoudi (approved):

PR #308, opened 2023-04-27
SQL API extensions to support joins

kbeedkar:
This PR makes extends the SQL API to also handle joins.

- Algebraic Joins are transpiled to Wayang Join Operator
- Since the JoinOperator outputs a Tuple2<Record, Record>, the join is fused with a MapOperator to output <Record>
- This is required to have downstream operations on joins such as projections/aggregations.

With this PR merged, Wayang can now support simple and basic SPJ queries for decoupled JDBC sources and CSV data sources.


zkaoudi (approved):
Great addition! Wayang can now accept SPJ SQL queries as input.

PR #310, opened 2023-04-30
Add cross-platform join

michellesackmann:
Add support for join across csv and postgres


2pk03 (approved):
build works, the license check should not fail. Thank you @michellesackmann!

PR #311, opened 2023-04-30
Updates for monitoring the runtime metrics of underlying platform (Spark)

AdeelAslamUnimore:
I have made some updates to enhance the monitoring of runtime metrics for the underlying platform. Initially, I implement for Spark big data platform, however, I will extend it for other underlying platform of Apache Wayang for debugging purpose  Platform-level runtime metrics monitoring allows for more granular resource utilization and performance metrics tracking during Spark job execution.

This will enable more effective monitoring and optimization of platform job performance, and provide greater visibility into resource utilization, allowing for more efficient resource allocation and troubleshooting.

Please let me know if you have any questions or feedback on these updates. Thank you for considering this pull request.


2pk03 (changes requested):
Hi @AdeelAslamUnimore - thank you. Please add the required Apache Software License headers in https://github.com/apache/incubator-wayang/actions/runs/4844948649/jobs/8633760143?pr=311#step:3:127

zkaoudi:
Hi all,

I'm wondering if modifying WayangContext and Job to include the hackIT variables/methods is the "best" way to go. Another alternative would be to create a new class eg. WayangContextWithMonitoring which extends from wayang context and contains whatever is necessary. Same for the Job class. 
What do you think? @JorgeQuiane @kbeedkar Any thoughts?

kbeedkar:
I second that. It's also how sqlcontext does. In general, we should avoid modifying Wayangcontext as this may have undesirable side effects.

zkaoudi:
If sqlcontext does the same (or some other module), then if someone needs to monitor jobs using the sql api, how would he do it? which context would he have to use? Just brainstorming here..

kbeedkar:
Good point. There is already a Monitor abstract class, and the current WayangContext provides an execute() method with Monitor. That would be a clean way IMO.

kbeedkar:
Hi guys, I went through the code. Overall, while the PR extends WayangContext, it does not affect any existing methods so this should not break anything. A few other issues with the code are
1. package name should start in lowercase, also there is a typo (Intefaces --> interfaces)
2. package names should be more meaningful and be consistent with the code base. spark_monitoring --> spark-monitoring. I would suggest having another meaningful name as the package is already under /spark/monitoring
3. Comments should not have @author tags (or?)

JorgeQuiane:
IMHO the monitoring should be part of the WayangContext as it is a core functionality and the SQLContext should extend from the WayangContext so that all (current and future) functionalities are inherited. As Kaustubh pointed out, we already have a Monitor abstract class and Monitor functionality in the current WayangContext. I also believe this should be the way to go.

zkaoudi:
Ok then! We keep the wayangcontext as is. 
@AdeelAslamUnimore can you please make the required changes that @kbeedkar? 
Then we can merge :)

BTW, I think there are no author tags in general in any class. I think it's an Apache decision.

2pk03 (commented):
+1

2pk03:
This PR is behind our current main, can you please rebase? We updated pom.xml to pass Eclipse licenses.

zkaoudi:
Hi @AdeelAslamUnimore, there are some compilation errors. Can you please take a look?
https://github.com/apache/incubator-wayang/actions/runs/4881291333/jobs/8710465408?pr=311

AdeelAslamUnimore:
Hi @zkaoudi The compilation problem occur is due to **spark_2.11**. Some definitions used in our SparkListener are defined only for wayang spark_2.12, i.e., executorInfo.resourceProfileId() is not present in spark_2.11. Similarly, for stageExecutorMetrics.

zkaoudi:
Hi @AdeelAslamUnimore, thank you for the explanation. The build is passing now. Did you change your implementation? Does the constraint of using spark_2.12 still hold?

AdeelAslamUnimore:
Hi @zkaoudi,
I have discovered that there are three conflicts for sparkListener with Wayang_Spark2.11 and Wayang_Spark2.12. Therefore, I am not utilizing the updated (2.12 SparkListener) attributes in my ownSparkListener for monitoring purposes. Now it should work for both releases.

zkaoudi:
Great! thank you! @2pk03 We can merge this :)

2pk03 (approved):

PR #313, opened 2023-05-02
Revert "Bump calcite-core from 1.29.0 to 1.32.0 in /wayang-api/wayang-api-sql"

zkaoudi:
Reverts apache/incubator-wayang#302 to avoid building errors

### DIFF
--- a/wayang-api/wayang-api-sql/pom.xml
+++ b/wayang-api/wayang-api-sql/pom.xml
@@ -33,7 +33,7 @@
         <dependency>
             <groupId>org.apache.calcite</groupId>
             <artifactId>calcite-core</artifactId>
-            <version>1.32.0</version>
+            <version>1.29.0</version>
         </dependency>
         <dependency>
             <groupId>org.apache.calcite</groupId>


2pk03:
No, its a license check issue. Pls read my comments as I merged.

zkaoudi:
I don't understand what's going on. 
Why now Wayang is not building with the previous version of calcite? It used to, right?

kbeedkar:
Calcite 1.29.0 had no license issues and should build.

zkaoudi:
Well now it's not building and gives the same error about forbidden licences for some strange reason.. Anyone has an idea on how to proceed?

2pk03:
I raised this question in the general incubator list, no response so far. You can try to revert the latest SQL patch, too, so solve this as a hotfix.

zkaoudi:
Great @2pk03! Let's hope they reply soon..

Which SQL patch you refer to?

2pk03:
https://github.com/apache/incubator-wayang/pull/310 eventually, that would be the state before bump up Calcite.

zkaoudi:
Ah that's a good point, there is an edit in the pom file.. I will check that. Thanks Alex!

zkaoudi:
Ok, now I understand: PR https://github.com/apache/incubator-wayang/pull/310 adds a new calcite dependency and this fails in the license checks as well. 
So we either revert that PR or we modify the main pom.xml to skip these checks.. 
Which one your prefer?
Both solutions are temporary until we figure out what is really the problem..

2pk03:
I'd say license check skip or adjust, the license in Calcite is legit, since it's a binary one, it seems. Maybe we can tune the license check [1] to be inline with the current Apache License compatibility [2]?
[1] http://www.mojohaus.org/license-maven-plugin/
[2] https://www.apache.org/legal/resolved.html

kmoltke:
Let me have a look if the SQL part works as intended if we leave out the `calcite-file` dependency. Then we can just remove it. Give me 5 min

kbeedkar:
Note that the calcite file is required for integrating file-based tables (e.g., CSV)

kmoltke:
+1 @kbeedkar  I also just tested that, and it is required..

kbeedkar:
> Ok, now I understand: PR #310 adds a new calcite dependency and this fails in the license checks as well. So we either revert that PR or we modify the main pom.xml to skip these checks.. Which one your prefer? Both solutions are temporary until we figure out what is really the problem..

I'd be in favor of reverting to 1.29.0 (for now) until we figure out a solution.

zkaoudi:
There is no problem with the licences now so I close this PR

PR #318, opened 2023-05-05
Support sql query on multiple csv table files

kmoltke:
A minor change to SQL API to allow to run queries on multiple csv files. It is still hard coded for now, but this commit is a tiny improvement.


wayang-api/wayang-api-sql/src/main/java/org/apache/wayang/api/sql/calcite/rel/WayangTableScan.java
```diff
@@ -78,7 +78,7 @@ public static WayangTableScan create(RelOptCluster cluster, RelOptTable relOptTa
 
     @Override
     public String toString() {
-        return "Wayang TableScan ["+ getQualifiedName() + "]";
+        return "Wayang TableScan [" + getQualifiedName() + "]";
     }
```
CalvinKirs, line 83: @2pk03 Do we have a unified code style check?

novatechflow: not yet, we should ;)

zkaoudi (changes requested):
Hi @kmoltke, thanks for fixing this. Is it possible to remove the hard-coded path? We wouldn't like to have hard-coded paths in the release ;)

kmoltke:
Hi, we spent several hours trying to do that. Unfortunately we had to prioritize finishing the report (we also agreed with Jorge about this). I believe it is possible though. Our problem was that it wasn't possible to access the file path through the RelNode. We were able to actually see the filesource during debugging, but the field was protected and not accessible. There is an issue on this: https://github.com/apache/incubator-wayang/issues/312 . Also, I think it would be good to create a new CsvTableScan class that extends the TableScan, so that the TableScan class will remain more generic (e.i. if source path is not required for a table scan). If we manage to somehow find time in the pretty packed exam period, we will definitely give it another go! :)

zkaoudi:
Maybe adding a get() method to access the path could work? Not sure if it makes sense in general though..

kmoltke:
Hi Zoi, we rearranged priorities with @JorgeQuiane and start working on getting rid of this hardcoded line. Just an fyi if it has an affect on the release plans :)

kmoltke:
Closed due to new PR. Please see https://github.com/apache/incubator-wayang/pull/322 instead.

PR #319, opened 2023-05-06
Updated Flink version and added some operator tests

damik3:
Support for Java 11 starts from Flink 1.10, whereas the project had Flink 1.7.1. See [here](https://nightlies.apache.org/flink/flink-docs-release-1.10/release-notes/flink-1.10.html#java-11-support-flink-10725).

I also, changed the Flink version on the wayang-tests-integration pom so as to have the same version throughout the project.


zkaoudi (approved):
Great! Thanks for spotting that @damik3 :)

PR #340, opened 2023-09-05
minor correction, added license header to some md files

2pk03:


chrisdutz:
As mentioned in Slack ... please use the official Apache Headers described here: https://www.apache.org/legal/src-headers.html and not the ones another podling made up ;-) (Only Pekko had these headers and they will most probably not have them for their next release ;-) ... so thanks for pointing this out ;-))

chrisdutz (changes requested):
Wrong Apache Headers

2pk03:
Changed to the correct headers.

zkaoudi (approved):

chrisdutz (approved):

PR #346, opened 2023-10-06
Master

AdityaGoel11:


AdityaGoel11:
Add feature: Support for basic SQL Aggregation queries for SUM, MIN, MAX, COUNT, and AVG aggregation functions

PR #348, opened 2023-10-10
Modified the JavaTextFileSource so that a remote file can be loaded via HTTPS protocol (or HTTP)

kamir:
**JavaTextFileSource.java** provides a selection based on the file URL's protocol.
If **https** or **http** is used instead of file, we use an HTTPConnection Object to load the data.

The test class is **JavaTextFileSourceTest.java**.

The test runs on my system in the branch **rel/0.7.1**.

PR #352, opened 2023-10-12
Modified the JavaTextFileSource so that a remote file can be loaded via HTTPS protocol (or HTTP) #348

kamir:
Second attempt - now to the right branch.


kamir:
I have to work with the right branch locally, before starting a new PR attempt.
Hence I closed this one, which is still based on the release branch.

PR #353, opened 2023-10-12
Modified the JavaTextFileSoModified the JavaTextFileSource so that a remote file can be loaded via HTTPS protocol (or HTTP) #348urce so that a remot file can be loaded vi…

kamir:
JavaTextFileSource.java provides a selection based on the file URL's protocol.
If https or http is used instead of file, we use an HTTPConnection Object to load the data.

The test class is JavaTextFileSourceTest.java.

The test runs on my system in the branch rel/0.7.1 and on my own branch based on main (v0.6.1).


kamir:
One test requires "a local HTTP Server which allows to GET the file in path http://localhost:8000/LICENSE.
This causes the test to fail in the automates test context.
I will comment out that test.
It only tests the HTTP Protocol instead of the HTTPS protocol used in the other test.

zkaoudi (approved):
All tests are passing now

Issue #342, opened 2023-09-20
fix scala mentions in release notes


pjfanning:
Scala releases include 2.11, 2.12 and more recently 2.13 and 3.

The Wayang release notes keep referring to Scala 11 and 12 when I think you mean 2.11 and 2.12.

Examples:
* https://lists.apache.org/thread/l6y7v3y1f76hrhblz5f1znptn0dqhh9n
* https://github.com/apache/incubator-wayang/blob/main/RELEASE_NOTES

PR #357, opened 2023-10-25
Update RELEASE_NOTES

2pk03:
fix issue #342

### DIFF
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -1,8 +1,6 @@
 ==============================================================
-Apache Wayang (incubating) 0.6.0
+Apache Wayang (incubating) 0.7.1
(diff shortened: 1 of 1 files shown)


RELEASE_NOTES
```diff
@@ -22,26 +20,9 @@ could be different.
...
 ------------
 
-support M1 ARM architecture
+Basic SQL support
 
 Incompatible changes
 --------------------
 
 - NONE -
-
```
glauesppen, line 31: Are we going to remove these lines?

novatechflow: Ya, since I didn't found anything in our current release notes. When I missed something please add :)

glauesppen (approved):

PR #365, opened 2023-10-31
Feature Update: Aggregation queries in SQL API

AdityaGoel11:
Add support for handling aggregation queries with an arbitrary number of aggregation and grouping columns.


AdityaGoel11:
Add support for projection queries with constants(RexLiterals) and ADD/SUBTRACT/DIVIDE/MULTIPLY operators in RexCalls.

kbeedkar (approved):
LGTM. However, going ahead we may have to rethink about generating Wayang operator UDFs.

PR #372, opened 2023-11-13
ML4all module

zkaoudi:
First versio of ML4all module. It contains the abstraction we discuss in the paper and two implementations, one for kmeans and one for SGD. More improvements to come.


kbeedkar (approved):
Thanks for the excellent contribution @zkaoudi!

Issue #384, opened 2023-12-05
Jdbc and Postgres do not support JoinOperators yet


juripetersen:
The Postgres Platform only supports simple select-project queries for now.
Introducing Joins would expand the possibilities of using Postgres with Wayang.

I will open a pull request and propose a solution to this.

kbeedkar:
At the moment only simple projections and filters can be pushed down. Supporting join push downs is welcomed.

zkaoudi:
+1

PR #385, opened 2023-12-05
Add JdbcJoinOperator and PostgresJoinOperator

juripetersen:
References #384


kbeedkar:
@juripetersen Thanks for the contributions. Could you please check couple of comments I left?

juripetersen:
> @juripetersen Thanks for the contributions. Could you please check couple of comments I left?

Sure, however I can't seem to find your comments yet.

juripetersen:
I can see the problem of having a cyclic reference and will remove that.

kbeedkar:
That makes sense. However, I suggest checking if we can follow the same approach as for filters.
https://github.com/apache/incubator-wayang/blob/5e6a07edaae3e7ebac9ac73814f95bafe741845d/wayang-platforms/wayang-postgres/src/main/java/org/apache/wayang/postgres/operators/PostgresFilterOperator.java#L30

i.e., the postgresJoin's constructor takes in two arguments of type TransformationDescriptor<Record, Key>

juripetersen:
@kbeedkar the `PostgresFilterOperator` uses the `sqlImplementation` and the executor only compiles this "hardcoded" information into the desired SQL query string. This can be seen in the JDBC `FunctionCompiler`:

```java
public class FunctionCompiler {

    /**
     * Compile a predicate to a SQL {@code WHERE} clause.
     *
     * @param descriptor describes the predicate
     * @return a compiled SQL {@code WHERE} clause
     */
    public String compile(PredicateDescriptor descriptor) {
        final String sqlImplementation = descriptor.getSqlImplementation();
        assert sqlImplementation != null;
        return sqlImplementation;
    }
}
```

This is also where the currently proposed solution of this PR was inspired from. @zkaoudi  and I were discussing an approach to avoid this and actually use a generic interface in Joins. However, this seems to be non-trivial and is likely to break backwards compatibility.

juripetersen:
I can imagine using something like this when defining the Join:

```java
        final ExecutionOperator joinOperator = new HsqldbJoinOperator<Integer>(
            new JoinDescriptor<Record, Record, Integer>(
                "tableName1",
                "tableName2",
                "keyName1",
                "keyName2",
                Record.class,
                Record.class,
                Integer.class
            )
        );
```

Where `JoinDescriptor` extends `TransformationDescriptor` and `JdbcJoinOperator` can only be derived from `JoinOperator`s that are defined with `JoinDescriptor`.

zkaoudi:
@juripetersen can you please check? There are some compilation errors that make the build fail.

kbeedkar:
Yes, extending the TransformationDescriptor sounds good.

juripetersen:
@zkaoudi @kbeedkar compilation works fine locally now.

zkaoudi (approved):

Issue #283, opened 2023-04-13
Code restructure


zkaoudi:
Currently, there are some modules that the source code is not in the usual path of "[module_name]/src/main/java" but inside a "code" folder. This is not the case for all modules. We should restructure the code to be as it is done usually: "[module_name]/src/main/java"

NOTE: It seems that this "code" folder was created because of some incompatibility with GraphChi and the java/scala versions. So we may have to discard GraphChi

2pk03:
+1 to retire GraphChi, we also should considering to retire Java 8 support.

bgeng777:
+1 for the restructure and retire GraphChi. If we want to take a further step after removing GraphChi, it is worthwhile to consider retiring Scala2.11 as well. 
The underlying engines have done such removal:
Spark dropped Scala2.11 since Spark3.0. See [here](https://issues.apache.org/jira/browse/SPARK-26132)
Flink dropped Scala2.11 since Flink1.15. See [here](https://issues.apache.org/jira/browse/FLINK-20845)
It can reduce matainers' bandwidth on different scala version. Feel free to correct me if I may miss some potential dependency on Scala2.11 in Wayang. :)

2pk03:
+1

jimtds:
One reason the ../code/.. directories exist is to avoid multiple Scala versions deps within the same project. So If it is needed to support multiple versions, maybe a strategy like 1 branch/version would be a good idea to avoid code duplication due to deps with different versions.

2pk03:
bump

zkaoudi:
What about retiring scala 2.11?

kbeedkar:
+1

juripetersen:
I will take care of implementing the restructure of this issue.

zkaoudi:
Great! Thanks Juri!

PR #389, opened 2024-01-10
Retire Java8/Scala 2.11 support in restructure

juripetersen:
This PR is supposed to implement the restructure necessary to retire Scala 2.11/Java 8 as mentioned in #283.

As of now, the PR doesn't retire GraphChi, but after removing the support for Scala 2.11 and Java 8, it seems like this is the next step or should be included in this PR as well. If so, just let me know in a comment here.


zkaoudi:
I vote to also remove graphchi in this PR.

zkaoudi (changes requested):
There something wrong with the build and some pom file I guess. @juripetersen can you please check it out?

zkaoudi (approved):
Great job!

zkaoudi:
This PR closes #283

PR #399, opened 2024-01-30
Update NOTICE

Apricity001:
Update copyright to year 2024 in NOTICE


2pk03 (approved):
Ha, good catch - thank you :)

PR #426, opened 2024-04-08
Minor changes for compilation with raw mvn

Baunsgaard:
When opening Wayang in my IDE it produced many errors, 

image

This PR fixes all of them (at least for my setup).

The main issue is that a default Java compile level is not set in the root POM.xml file, and therefore my IDE(vscode) does not know what to compile. While it is possible to add this as a configuration inside an individual's setup .vscode for Visual Code, I would still suggest moving it to a default setting. Because 1. the .vscode settings file is git ignored and 2. the IDE automatically picks up the default value set in the root parameters of the POM.xml.

The second issue i had was with wayang-api/wayang-api-scala-java/pom.xml. 
Unfortunately, my IDE did not compile correctly without adding a recursive dependency on its package:

```
<dependency>
    <groupId>org.apache.wayang</groupId>
    <artifactId>wayang-api-scala-java_2.11</artifactId>
    <version>0.7.1</version>
</dependency>
```

Maybe there is a better way to handle this?

Finally, there were two minor bugs: 1. the docs pom contained the -SNAPSHOT when referencing the root repository, and 2. some imports contained an extra `;`.

Afterwards:

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,6 @@ pom.xml.*
 # VSCode
 .vscode
 .vscode/*
+
+# Scala Plugin for VSCode
+.metals
--- a/pom.xml
+++ b/pom.xml
@@ -106,6 +106,15 @@
         <scala.mayor.version>2.12</scala.mayor.version>
         <spark.version>3.1.2</spark.version>
 
+        <java.version>11</java.version>
+        <source.level>11</source.level>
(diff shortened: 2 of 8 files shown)


zkaoudi:
Thanks a lot for this PR Sebastian.

zkaoudi (approved):

PR #453, opened 2024-05-28
Set logLevel on WayangContext to allow doing it without conf file

juripetersen:
This proposes a functional interface to change the Loggers level with the `WayangContext`:

```java
context.setLogLevel(Level.ERROR);
```

The change provides better ergonomics and prevents having to fiddle with configuration files, when Context objects etc. can be created during plan building. Furthermore, this aligns with the API presented in other Apache projects like [Spark](https://github.com/apache/spark/blob/df5ef91f44bacf60b39bccecc049eb8cb5714b39/core/src/main/scala/org/apache/spark/SparkContext.scala#L398)

### DIFF
--- a/wayang-commons/wayang-core/src/main/java/org/apache/wayang/core/api/WayangContext.java
+++ b/wayang-commons/wayang-core/src/main/java/org/apache/wayang/core/api/WayangContext.java
@@ -30,6 +30,9 @@ import org.apache.wayang.core.util.ExplainUtils;
 import org.apache.wayang.core.util.ReflectionUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.apache.logging.log4j.Level;
+import org.apache.logging.log4j.core.LoggerContext;
+import org.apache.logging.log4j.core.config.LoggerConfig;
 
 /**
  * This is the entry point for users to work with Wayang.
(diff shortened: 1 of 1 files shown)


kbeedkar:
Can you check why some unit tests are failing?

juripetersen:
> Can you check why some unit tests are failing?

I can't replicate the error locally. It looks like building the assembly fails due to local/cache problems? Maybe a rerun of the action solves it?

juripetersen:
@kbeedkar it seems like the build can't find `wayang-api-json` in https://repo.maven.apache.org/maven2/org/apache/wayang/. This seems to be correct, as that module is not to be found there. Do we have to publish a release for that module to be in maven-central?

2pk03:
@juripetersen - yes I think so.

zkaoudi:
But the change does not concern the missing module, so that is weird. Also the module is locally located.

juripetersen:
@zkaoudi @kbeedkar I can reproduce this by running `./mvnw clean verify -B -Dmaven.test.skip=false -Dwayang.configuration=file:$(pwd)/tools/test/config/wayang.properties` on a fresh setup. The error occurs because some modules are not on mvn central as they are not in a release yet. I think this is sensible and the build should be constructed from the current state of the repo and not from central. If I run `mvn clean install` before, the local modules are built and then used for the verification step. This resolves the error. 

Is including install as a step before verification in our action a solution, or does this introduce something I am unaware of?

zkaoudi:
@juripetersen makes totally sense. @2pk03 do you see any issue with that?

PR #473, opened 2024-09-21
added the demo cluster configuration for Apache Wayang Demos

kamir:
The first iteration on creating a PLAYGROUND for Apache Wayang demos has been created.


juripetersen:
@kamir there are a lot of setup-specific configurations in this such as docker files, your personal paths in configurations etc.
Do you think this PR is worth updating or should we close it?

PR #480, opened 2024-11-04
Make Wayang work in Docker

avborup:
Devcontainer and more complete Dockerfile is yet to be added.


avborup:
Mistakenly opened a PR to the upstream fork rather than my own fork. My bad.

PR #481, opened 2024-11-13
Add /docs  to serve maven documentation for Wayang

2pk03:


2pk03:
Ya we need to exclude the directory

zkaoudi:
Same error for the license when building with the .js files

2pk03:
of course - the license check change is in the PR ;) It's just docs, approving should not be a problem ...

zkaoudi (approved):

PR #487, opened 2024-11-26
Update .asf.yaml

2pk03:
changed to better descriptions and tags for the project

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -16,23 +16,13 @@
 #
 
 github:
-  description: Apache Wayang(incubating) is the first cross-platform data processing system.
-  homepage: https://wayang.incubator.apache.org/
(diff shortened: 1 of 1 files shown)


zkaoudi (commented):
Let's not put federated learning in the title until we have the code finalized

kamir:
Until we have the code for federated learning, we could use:

"**Building An Industrial Data Federation Framework for Agents**"

as a step towards the proposed phrase: "Building An Industrial Federated Learning Framework"

mspruc:
Can we not merge this when @zkaoudi still has outstanding issues with the commit? Personally I also have issues with mentioning federated learning, when it seems like we have no capabilities and it's steering the project in a certain direction without broader approval.

PR #496, opened 2025-01-27
update tensorflow java to 1.0.0

joker-star-l:

PR #497, opened 2025-01-31
Revert "update tensorflow java to 1.0.0"

zkaoudi:
Reverts apache/incubator-wayang#496


zkaoudi:
No need to revert it, seems a local problem.

PR #519, opened 2025-03-26
[CI] support parallel builds with JDK 11 and JDK 17

CalvinKirs:
- support parallel builds with JDK 11 and JDK 17
- replace groovy-maven-plugin with gmavenplus-plugin, as groovy-maven-plugin is no longer actively maintained and does not support JDK 17.

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -32,25 +32,28 @@ jobs:
   check-license:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
       - name: Check License Header
         uses: apache/skywalking-eyes/header@main
   build:
     name: Build
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        java: [11, 17]
     steps:
-      - uses: actions/checkout@v2
-      - name: Set up JDK 11
+      - uses: actions/checkout@v4
+      - name: Set up JDK ${{ matrix.java }}
         uses: actions/setup-java@v2
         with:
-          java-version:  11
-          distribution: 'adopt'      
+          java-version: ${{ matrix.java }}
+          distribution: 'adopt'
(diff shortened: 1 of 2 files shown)


juripetersen:
Thank you for your contribution!
Right now, the unit tests are failing due to Mockito having problems in some of the tests. 
Did you verify your changes locally?


Another question that arises for me: 
By enabling builds with JDK 17, we enable a newer version of Java. Did you confirm that this can be handled with the current version of platforms like spark and flink used?

zkaoudi:
Hi @CalvinKirs , our project is not ready yet to support java 17 but we are close to make the right fixes. But I am still wondering if we have tests that use Java 17 features how would the build with Java 11 work? Or we shall refrain from having such tests to ensure backward compatibility?

CalvinKirs:
> Hi @CalvinKirs , our project is not ready yet to support java 17 but we are close to make the right fixes.
oops...my email may not have been officially sent... In fact, I noticed that our mailing list is considering upgrading to JDK 17, so I ran some tests through CI. Simply put, compilation works fine, but there are some compatibility issues when running tests. I haven’t looked into the details, but it might require upgrading some dependencies.

> But I am still wondering if we have tests that use Java 17 features how would the build with Java 11 work? Or we shall refrain from having such tests to ensure backward compatibility?
 Actually, we can build and test in parallel—kind of like this: we compile and test using both JDK 11 and 17 at the same time.

zkaoudi:
Yes we are working on using JDK 17 but not pushed changes yet.
Can we define different tests for JDK 11 and JDK 17? 

Also another question that may be worth discussin in the mailing list is do we want to keep jdk 11 or we simply assume only jdk 17?

novatechflow:
@CalvinKirs @zkaoudi - any updated or close this one?

zkaoudi:
we don't support java 11 anymore. I close this.

PR #521, opened 2025-03-31
Fix join key extraction for python join operator

juripetersen:
The python join operator's key extraction didn't work properly due to incorrect input handling. 
This resulted in a cartesian product, as both keys were null.

### DIFF
--- a/python/src/pywy/core/serializer.py
+++ b/python/src/pywy/core/serializer.py
@@ -53,12 +53,13 @@ class JSONSerializer:
 
         json_operator["data"] = {}
 
-        if hasattr(operator, "input_type"):
-            if operator.input_type is not None:
-                json_operator["data"]["inputType"] = ndim_from_type(operator.input_type).to_json()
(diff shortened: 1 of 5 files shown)


zkaoudi (changes requested):
Can you check with the  SparkIntegrationIT.testKMeans test fails?

juripetersen:
> Can you check with the SparkIntegrationIT.testKMeans test fails?

This test failed for the previous pull requests as well. It's likely an issue introduced in the 1.0.1 version and has nothing to do with this PR.

zkaoudi:
@joker-star-l Maybe you can also help as it's the sparkML that fails for some reason..

joker-star-l:
> @joker-star-l Maybe you can also help as it's the sparkML that fails for some reason..

image

This is caused by dependency conflicts.

zkaoudi:
Thank you Mingxi, indeed it was a dependency issue.

zkaoudi (approved):

PR #522, opened 2025-04-01
Fix spark dependency version to fix integration tests

juripetersen:
The verify step ran through locally.

The org.codehaus.janino dependency of spark needed to be set to a desired version.


zkaoudi (approved):
Great, thanks!

PR #531, opened 2025-04-03
add support for cross joins in the sql-api

mspruc:
Added support & tests for cross joins / cartesian products in the sql-api, still missing an implementation for the ```JdbcExecutor```, so it just works for serialisation-based / java platforms.


juripetersen (approved):

wayang-api/wayang-api-sql/src/main/java/org/apache/wayang/api/sql/calcite/converter/functions/FlattenJoinResult.java
```diff
@@ -0,0 +1,31 @@
+package org.apache.wayang.api.sql.calcite.converter.functions;
```
juripetersen, line 19: The license header for this file is missing

juripetersen (changes requested):

juripetersen (approved):

PR #533, opened 2025-04-06
Update Parquet operators to better use metadata early on

aurb9:
Users no longer need to provide the column names, they will be read by the operator itself.


aurb9:
Are RecordType and Record linked somehow? We might want to be able to perform record.getField(name) instead of record.getField(index), if a RecordType is used. This would facilitate the use of this operator very much...

zkaoudi (approved):

PR #558, opened 2025-05-02
Serialisation for aggregation in sql-api & improvements to various tests

mspruc:


juripetersen (approved):

2pk03 (approved):
+1 - one idea maybe, for uncheckedIOException could we add a explanation? Like permissions, file / table not found etc? Just an idea, no blocker!

mspruc:
@2pk03 you mean the one in `JdbcExecutor`:
[https://github.com/apache/incubator-wayang/blob/1788e1b3e465685a08a0a266dfc5ca8af62cd8d8/wayang-platforms/wayang-jdbc-template/src/main/java/org/apache/wayang/jdbc/execution/JdbcExecutor.java#L274]()?

This just seems like dead code, I dunno that it has any usage either tbh.

2pk03:
Ah, good, so we remove it in one of the next commits.

PR #562, opened 2025-05-07
Add explain to json config to Job

juripetersen:


wayang-commons/wayang-core/src/main/java/org/apache/wayang/core/api/Job.java
```diff
@@ -304,6 +305,16 @@ protected void doExecute() {
...
             }
 
+            if (this.configuration.getBooleanProperty("wayang.core.explain.enabled")) {
+                ExplainUtils.write(
+                    ExplainUtils.parsePlan(this.wayangPlan, true),
+                    this.configuration.getStringProperty("wayang.core.explain.logical.file")
+                );
+                ExplainUtils.write(
+                    ExplainUtils.parsePlan(executionPlan, true),
+                    this.configuration.getStringProperty("wayang.core.explain.execution.file")
```
zkaoudi, line 315: what if the user did not specify the file path?

zkaoudi (approved):

PR #568, opened 2025-05-15
Spark & Calcite Guava dependency issue

### LINKED ISSUE #567: Misalignment of calcite-version in pom.xml and wayang-assembly/pom.xml
The two versions are not aligned, leading to the project being compiled with 1.39 but execution through the assembly will run on 1.35. Thus, `SqlFunctions.LikeFunction` and other do not exist and will lead to an Exception.

Proposed fix:

Align both to 1.36 and use the old `SqlFunctions.like()` as before. This should resolve the conflicts.

mspruc:
Closes #567


mspruc:
@juripetersen Can I get you to run your configuration on this PR and see if it works? Thanks :).

juripetersen (approved):

PR #570, opened 2025-05-16
Add LogisticRegressionOperator with API training support and integration test

xristlamp:
This pull request introduces support for training logistic regression models in Apache Wayang via the public API.

Summary of changes:
- Implemented `LogisticRegressionOperator` (logical operator).
- Added `SparkLogisticRegressionOperator` for Spark execution.
- Defined `LogisticRegressionModel` interface.
- Created `LogisticRegressionMapping` to map the logical operator to Spark.
- Registered the mapping in Spark's ML plugin.
- Extended the public API:
  - Added `.trainLogisticRegression(...)` to `DataQuanta` and `DataQuantaBuilder`.
  - Introduced `LogisticRegressionDataQuantaBuilder` class (consistent with `dlTraining`).

-  Added an integration test: `testLogisticRegressionWithAPI` in `SparkIntegrationIT.java` that verifies:
  - `.loadCollection(...)`
  - `.trainLogisticRegression(...)`
  - `.predict(...)`
  - `.collect()`

Testing:
- Verified with `mvn clean install`: all tests pass successfully.

### DIFF
--- a/python/src/pywy/basic/model/models.py
+++ b/python/src/pywy/basic/model/models.py
@@ -32,5 +32,4 @@ class LogisticRegression(Op):
     def __init__(self, name=None):
         super().__init__(Op.DType.FLOAT32, name)
 
-    def inputs_required(self):
-        return 2
+


wayang-api/wayang-api-scala-java/src/main/scala/org/apache/wayang/api/DataQuanta.scala
```diff
@@ -105,6 +105,17 @@ class DataQuanta[Out: ClassTag](val operator: ElementaryOperator, outputIndex: I
                                       udfLoad: LoadProfileEstimator = null): DataQuanta[NewOut] =
     mapPartitionsJava(toSerializablePartitionFunction(udf), selectivity, udfLoad)
 
+
+  def trainLogisticRegression(labels: DataQuanta[java.lang.Double], fitIntercept: Boolean): DataQuanta[LogisticRegressionModel] = {
+    val operator = new LogisticRegressionOperator(fitIntercept)
+    this.connectTo(operator, 0)
+    labels.connectTo(operator, 1)
+    new DataQuanta[LogisticRegressionModel](operator)
```
zkaoudi: Why create a new DataQuanta and not just output the operator as it's done for the other cases (see dlTrainingJava)?

python/src/pywy/basic/model/ops.py
```diff
@@ -221,6 +221,13 @@ def get_dType(self):
     def inputs_required(self):
         return 1
 
+class LogisticRegression(Op):
```
zkaoudi: Not sure that's the right place for this. Checkout this package: 
from pywy.basic.model.models

zkaoudi (changes requested):

python/src/pywy/basic/model/models.py
```diff
@@ -27,3 +27,10 @@ def __init__(self, out: Op):
 
     def get_out(self):
         return self.out
+
+class LogisticRegression(Op):
+    def __init__(self, name=None):
+        super().__init__(Op.DType.FLOAT32, name)
+
+    def inputs_required(self):
```
zkaoudi: do we use this function somewhere?

xristlamp: You're right ! I removed the inputs_required method from the LogisticRegression class and re-ran the tests. Everything works perfectly without it, so it seems it's not needed in this case. Thanks for pointing it out!

zkaoudi (approved):

PR #579, opened 2025-05-24
ConvLSTM demo

joker-star-l:
Add ConvLSTM Demo (in [`TensorflowConvLSTMIT`](https://github.com/apache/incubator-wayang/pull/579/files#diff-87c4646b8251fda8be7493fec8ebf38084845cdd00e490fa380ed6797e619638))
The difference between this PR and the Python Pytorch version:
1. This PR uses the standard ConvLSTM2D, so that we can provide a built-in standard ConvLSTM2D layer. While the Python Pytorch version uses a user-defined ConvLSTM2D, which has little difference from the standard ConvLSTM2D.
2. This PR does not use BatchNorm because there are some bugs when building the computation graph. BatchNorm  has different behaviors in training and inference, which makes the computation graph complicated. I don't know how to solve the problem and maybe someone can fix it.
3. This PR does not use conv3D in the last layer, because conv3D in Tensorflow Java does not support backpropagation. Maybe this is a bug in Tensorflow Java.

### DIFF
--- a/wayang-platforms/wayang-tensorflow/src/main/java/org/apache/wayang/tensorflow/model/op/nn/TensorflowConv2D.java
+++ b/wayang-platforms/wayang-tensorflow/src/main/java/org/apache/wayang/tensorflow/model/op/nn/TensorflowConv2D.java
@@ -66,7 +66,7 @@ public class TensorflowConv2D<T extends TNumber> {
         throw new RuntimeException("Unsupported Stride: " + Arrays.toString(stride));
     }
 
-    public Operand<T> call(Operand<T> input) {
+    public Operand<T> callV1(Operand<T> input) {
         if (!op.getBias()) {
             return tf.withName(op.getName()).nn.conv2d(
                     input,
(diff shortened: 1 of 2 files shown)


zkaoudi:
Hi @joker-star-l, can you check the error that appears in the json api? 
https://github.com/apache/incubator-wayang/actions/runs/15519196678/job/44058594910?pr=579

zkaoudi:
Sorry, it's not the json api. 
Here is the errors:

Error:  /home/runner/work/incubator-wayang/incubator-wayang/wayang-api/wayang-api-json/src/main/scala/operatorfromjson/binary/DLTrainingOperatorFromJson.scala:46: error: type mismatch;
Error:   found   : org.apache.wayang.basic.model.op.Input.Type
Error:   required: Array[Int]
Error:        case "Input" => new Input(parseInputType(opType))
Error:                                                ^
Error:  /home/runner/work/incubator-wayang/incubator-wayang/wayang-api/wayang-api-json/src/main/scala/operatorfromjson/binary/DLTrainingOperatorFromJson.scala:77: error: value PREDICTED is not a member of object org.apache.wayang.basic.model.op.Input.Type
Error:        case "..PREDICTED.." => Input.Type.PREDICTED
Error:                                           ^
Error: [ERROR] two errors found

joker-star-l:
> Sorry, it's not the json api. Here is the errors:
> 
> Error: /home/runner/work/incubator-wayang/incubator-wayang/wayang-api/wayang-api-json/src/main/scala/operatorfromjson/binary/DLTrainingOperatorFromJson.scala:46: error: type mismatch; Error: found : org.apache.wayang.basic.model.op.Input.Type Error: required: Array[Int] Error: case "Input" => new Input(parseInputType(opType)) Error: ^ Error: /home/runner/work/incubator-wayang/incubator-wayang/wayang-api/wayang-api-json/src/main/scala/operatorfromjson/binary/DLTrainingOperatorFromJson.scala:77: error: value PREDICTED is not a member of object org.apache.wayang.basic.model.op.Input.Type Error: case "..PREDICTED.." => Input.Type.PREDICTED Error: ^ Error: [ERROR] two errors found

It seems that the CI failed because I changed the Java API for creating models. I tried to fix it temporarily. @zkaoudi Could you help me trigger the CI?

But maybe the json api and python api need to be modified simultaneously.

zkaoudi:
Another error occured now

joker-star-l:
> Another error occured now

This is a strange error about the data format in tensorflow conv2d operator on linux. This maybe is a bug for tensorflow. I use another data format to fix it.

 @zkaoudi Could you help me trigger the CI again?

zkaoudi (approved):

PR #580, opened 2025-05-27
Add Python API for DecisionTreeRegression

xristlamp:
This PR adds support for training Decision Tree Regression models in the Python API.

- Added DecisionTreeRegression operator in models.py
- Added .train_decision_tree_regression() method to DataQuanta
- Includes a test that trains the model and validates predictions


mspruc (approved):
Looks good!

PR #584, opened 2025-06-12
Parallelism for java textfilesources & code quality

mspruc:


mspruc:
after testing this has no performance gain, closing.

PR #587, opened 2025-06-17
Add Filter into join rule to avoid cartesian products

juripetersen:
This PR adds two rules to the SqlContexts main function, allowing to push down filters when possible. 

This avoids Cartesian products and produces the more expected Join with a condition.


mspruc (approved):
We probably add some way for the user to interact with the rules based optimisation from calcite, but for now this is good.

Issue #501, opened 2025-02-06
Minor edits in LICENSE/NOTICE files


zkaoudi:
- The LICENSE in the source distribution should mention that
Unsigned16 and Random16 are copied from Hadoop (in wayang-benchmark).
- apache-wayang-utils-profile-db-1.0.0-incubating.jar META-INF/LICENSE
and META-INF/NOTICE should mention code copied from profiledb. The
other jars are OK

pjfanning:
@zkaoudi would it be possible to do this before graduation?

I have added #604 for another unresolved issue that came up during your last release vote.

zkaoudi:
Yes! We will do these before graduating. They slipped somehow. Thank you PJ

PR #609, opened 2025-09-12
deleted unused and copied terasort classes

zkaoudi:
These classes were copied from Hadoop but we don't use them anywhere so it's better to delete them. This would solve half of issue #501


zkaoudi:
Closing because there are used in the scala apps

PR #638, opened 2025-12-08
Ported object-file IO onto a safe codec to mitigate gadget chain attack

novatechflow:
Port object-file IO onto a safe codec
------------------------------------

- add `ObjectFileSerializationMode` + helper to switch between legacy Java serialization and a JSON-based format, with unit tests covering both paths
- rework Java/Flink/Spark object-file sources/sinks (and Flink’s output format) to use the shared serializer, log deprecation warnings for the legacy mode, and stop calling raw `ObjectInputStream`
- update Spark’s SequenceFile readers/writers to chunk records via RDD transformations and share logic with the new serializer

Motivation: the legacy path deserializes untrusted SequenceFile payloads via `ObjectInputStream.readObject`, letting an attacker ship a gadget chain that runs arbitrary bytecode inside the Wayang JVM. That RCE can be used to execute system commands, exfiltrate data, or tamper with jobs, so we move to JSON by default and require explicit opt-in for the old codec.

========================== IMPORTANT =========================
Before we merge test the patch, ideally in a setup with source and sink platforms. Compilation and tests when through in my setup.

```
mvn -pl wayang-commons/wayang-basic -Dtest=org.apache.wayang.basic.operators.ObjectFileSerializationTest test
[INFO] Building Wayang Basic 1.1.1-SNAPSHOT
[WARNING] 2 problems were encountered while building the effective model for org.apache.yetus:audience-annotations:jar:0.5.0 during dependency collection step for project (use -X to see details)
[INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  2.683 s
[INFO] Finished at: 2025-12-08T19:47:45+01:00
[INFO] ------------------------------------------------------------------------
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@
   - under the License.
   -->
 
-# Apache Wayang (incubating) <img align="right" width="128px" src="https://wayang.apache.org/img/wayang.png" alt="Wayang Logo">
+# Apache Wayang™ <img align="right" width="128px" src="https://wayang.apache.org/img/wayang.png" alt="Wayang Logo">
 
 ## The first open-source cross-platform data processing system
 
@@ -48,9 +48,9 @@
 
 ## Description
 
-In contrast to traditional data processing systems that provide one dedicated execution engine, Apache Wayang (incubating) can transparently and seamlessly integrate multiple execution engines and use them to perform a single task. We call this *cross-platform data processing*. In Wayang, users can specify any data processing application using one of Wayang's APIs and then Wayang will choose the data processing platform(s), e.g., Postgres or Apache Spark, that best fits the application. Finally, Wayang will perform the execution, thereby hiding the different platform-specific APIs and coordinating inter-platform communication.
+In contrast to traditional data processing systems that provide one dedicated execution engine, Apache Wayang can transparently and seamlessly integrate multiple execution engines and use them to perform a single task. We call this *cross-platform data processing*. In Wayang, users can specify any data processing application using one of Wayang's APIs and then Wayang can choose the data processing platform(s), e.g., Postgres or Apache Spark, that best fits the application. Finally, Wayang will orchestrate the execution, thereby hiding the different platform-specific APIs and coordinating inter-platform communication.
 
-Apache Wayang (incubating) aims at freeing data engineers and software developers from the burden of learning all different data processing systems, their APIs, strengths and weaknesses; the intricacies of coordinating and integrating different processing platforms; and the inflexibility when trying a fixed set of processing platforms. As of now, Wayang has built-in support for the following processing platforms:
+Apache Wayang aims at freeing data engineers and software developers from the burden of learning all different data processing systems, their APIs, strengths and weaknesses; the intricacies of coordinating and integrating different processing platforms; and the inflexibility when trying a fixed set of processing platforms. As of now, Wayang has built-in support for the following processing platforms:
 - [Java Streams](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html)
 - [Apache Spark](https://spark.apache.org/)
 - [Apache Flink](https://flink.apache.org/)
@@ -65,7 +65,9 @@ Apache Wayang (incubating) can be used via the following APIs:
 - Java scala-like
 - Scala
 - SQL
-- Java native (recommended only for low level)
+- Java native (recommended only for low level development)
+
+Apache Wayang provides a flexible architecture which enables easy addition of new operators and data processing platforms without requiring any change of the internals of the system. For details on how to add new operators, see [here](https://wayang.apache.org/docs/guide/adding-operators).
 
 ## Quick Guide for Running Wayang
 


zkaoudi:
There are individual operator tests for sources and sinks in each platform. See eg: 
https://github.com/apache/incubator-wayang/blob/afb8e413c13b20c5094c818ec2d0b9cf92d499f0/wayang-platforms/wayang-java/src/test/java/org/apache/wayang/java/operators/JavaObjectFileSourceTest.java

Which setup would you suggest for testing besides these tests?

novatechflow:
@zkaoudi - did a local test with Flink, no issues - let’s merge if no objections. Can you please check if we might have a collision with #639?

wayang-commons/wayang-basic/src/main/java/org/apache/wayang/basic/operators/ObjectFileSink.java
```diff
@@ -36,6 +36,8 @@ public class ObjectFileSink<T> extends UnarySink<T> {
 
   protected final Class<T> tClass;
 
+  private ObjectFileSerializationMode serializationMode = ObjectFileSerializationMode.LEGACY_JAVA_SERIALIZATION;
```
juripetersen: If deprecated, shouldn't this utilize the JSON variant by default?

wayang-commons/wayang-basic/src/main/java/org/apache/wayang/basic/operators/ObjectFileSource.java
```diff
@@ -43,6 +44,8 @@ public class ObjectFileSource<T> extends UnarySource<T> {
 
     private final Class<T> tClass;
 
+    private ObjectFileSerializationMode serializationMode = ObjectFileSerializationMode.LEGACY_JAVA_SERIALIZATION;
```
juripetersen: If deprecated, shouldn't this utilize the JSON variant by default?

wayang-platforms/wayang-flink/src/main/java/org/apache/wayang/flink/compiler/WayangFileOutputFormat.java
```diff
@@ -125,6 +125,7 @@ public static void initDefaultsFromConfiguration(Configuration configuration) {
 
     private transient DataOutputViewStreamWrapper outView;
 
+    private ObjectFileSerializationMode serializationMode = ObjectFileSerializationMode.LEGACY_JAVA_SERIALIZATION;
```
juripetersen: If deprecated, shouldn't this utilize the JSON variant by default?

novatechflow: we should not switch off functions directly, instead give users a certain amount of time to use the new feature for backwards compatibility. Thoughts?

wayang-platforms/wayang-flink/src/main/java/org/apache/wayang/flink/operators/FlinkObjectFileSink.java
```diff
@@ -80,8 +88,14 @@ public Tuple<Collection<ExecutionLineageNode>, Collection<ChannelInstance>> eval
 
         //TODO: remove the set parallelism 1
         DataSetChannel.Instance input = (DataSetChannel.Instance) inputs[0];
+        ObjectFileSerializationMode serializationMode = this.getSerializationMode();
+        if (serializationMode == ObjectFileSerializationMode.LEGACY_JAVA_SERIALIZATION) {
```
juripetersen: This is repeated in every implementation extending ObjectFileSink, so it should just be moved there and called here for checking the serializationMode

wayang-platforms/wayang-flink/src/main/java/org/apache/wayang/flink/operators/FlinkObjectFileSink.java
```diff
@@ -47,6 +51,10 @@
  */
 public class FlinkObjectFileSink<Type> extends ObjectFileSink<Type> implements FlinkExecutionOperator {
 
+    private static final Logger LOGGER = LogManager.getLogger(FlinkObjectFileSink.class);
+
+    private static final AtomicBoolean LEGACY_WARNING_EMITTED = new AtomicBoolean(false);
```
juripetersen: This field could also be moved to the ObjectFileSink to make it less redundant.

wayang-platforms/wayang-flink/src/main/java/org/apache/wayang/flink/compiler/WayangFileOutputFormat.java
juripetersen: I think it depends on the consequences for the user. If they have to implement further things to make sure their data types implement Serializable, I agree. If this is a change without any overhead for the user, fixing a potential security threat, then it should be the new default.

wayang-platforms/wayang-flink/src/main/java/org/apache/wayang/flink/operators/FlinkObjectFileSink.java
novatechflow: Yeah, but (I know ...) from past experiences it's better to tell every time a soon deprecated function is called. We can refactor it, main goal is to close the already known security bug.

novatechflow:
created #640 - my fork gots messy with merges.

PR #640, opened 2025-12-09
Ported object-file IO onto a safe codec to mitigate gadget chain attack

novatechflow:
Port object-file IO onto a safe codec
------------------------------------

- add `ObjectFileSerializationMode` + helper to switch between legacy Java serialization and a JSON-based format, with unit tests covering both paths
- rework Java/Flink/Spark object-file sources/sinks (and Flink’s output format) to use the shared serializer, log deprecation warnings for the legacy mode, and stop calling raw `ObjectInputStream`
- update Spark’s SequenceFile readers/writers to chunk records via RDD transformations and share logic with the new serializer

Motivation: the legacy path deserializes untrusted SequenceFile payloads via `ObjectInputStream.readObject`, letting an attacker ship a gadget chain that runs arbitrary bytecode inside the Wayang JVM. That RCE can be used to execute system commands, exfiltrate data, or tamper with jobs, so we move to JSON by default and require explicit opt-in for the old codec.

Replaces PR #638

### DIFF
--- a/wayang-api/wayang-api-python/pom.xml
+++ b/wayang-api/wayang-api-python/pom.xml
@@ -37,7 +37,6 @@
         <spark.version>2.4.0</spark.version>
         <scala.mayor.version>2.12</scala.mayor.version>
         <giraph.version>1.2.0-hadoop2</giraph.version>
-        <python.worker.tests.skip>true</python.worker.tests.skip>
     </properties>
 
     <dependencyManagement>
--- a/wayang-platforms/wayang-spark/src/main/java/org/apache/wayang/spark/operators/SparkObjectFileSink.java
+++ b/wayang-platforms/wayang-spark/src/main/java/org/apache/wayang/spark/operators/SparkObjectFileSink.java
@@ -130,8 +130,8 @@ public class SparkObjectFileSink<T> extends ObjectFileSink<T> implements SparkEx
                                                                     ObjectFileSerializationMode mode) {
         try {
             byte[] payload = ObjectFileSerialization.serializeChunk(buffer, validLength, mode);
-            BytesWritable writable = new BytesWritable(payload);
-            return new Tuple2<>(NullWritable.get(), writable);
+
+            return new Tuple2<>(NullWritable.get(), new BytesWritable(payload));
         } catch (IOException e) {
             throw new UncheckedIOException("Failed to serialize Spark object file chunk.", e);
         }
--- a/wayang-platforms/wayang-spark/src/main/java/org/apache/wayang/spark/operators/SparkObjectFileSource.java
+++ b/wayang-platforms/wayang-spark/src/main/java/org/apache/wayang/spark/operators/SparkObjectFileSource.java
@@ -88,17 +88,13 @@ public class SparkObjectFileSource<T> extends ObjectFileSource<T> implements Spa
 
         final String actualInputPath = FileSystems.findActualSingleInputPath(sourcePath);
         final ObjectFileSerializationMode serializationMode = this.getSerializationMode();
-        if (serializationMode == ObjectFileSerializationMode.LEGACY_JAVA_SERIALIZATION) {
-            // Warning is emitted by ObjectFileSource#getSerializationMode.
-        }
         final JavaPairRDD<NullWritable, BytesWritable> rawRdd =
                 sparkExecutor.sc.sequenceFile(actualInputPath, NullWritable.class, BytesWritable.class);
         final Class<T> typeClass = this.getTypeClass();
         final JavaRDD<Object> rdd = rawRdd.flatMap(tuple -> {
             byte[] payload = Arrays.copyOf(tuple._2.getBytes(), tuple._2.getLength());
             try {
-                List<Object> chunk = ObjectFileSerialization.deserializeChunk(payload, serializationMode, typeClass);
-                return chunk.iterator();
+                return ObjectFileSerialization.deserializeChunk(payload, serializationMode, typeClass).iterator();
             } catch (IOException | ClassNotFoundException e) {
                 throw new UncheckedIOException(new IOException("Failed to deserialize Spark object file chunk.", e));
             }


wayang-platforms/wayang-spark/src/main/java/org/apache/wayang/spark/operators/SparkObjectFileSource.java
```diff
@@ -79,7 +87,22 @@ public Tuple<Collection<ExecutionLineageNode>, Collection<ChannelInstance>> eval
         RddChannel.Instance output = (RddChannel.Instance) outputs[0];
 
         final String actualInputPath = FileSystems.findActualSingleInputPath(sourcePath);
-        final JavaRDD<Object> rdd = sparkExecutor.sc.objectFile(actualInputPath);
+        final ObjectFileSerializationMode serializationMode = this.getSerializationMode();
+        if (serializationMode == ObjectFileSerializationMode.LEGACY_JAVA_SERIALIZATION) {
+            // Warning is emitted by ObjectFileSource#getSerializationMode.
```
juripetersen: This essentially does nothing, so it should just be removed.

wayang-platforms/wayang-spark/src/main/java/org/apache/wayang/spark/operators/SparkObjectFileSource.java
```diff
@@ -79,7 +87,22 @@ public Tuple<Collection<ExecutionLineageNode>, Collection<ChannelInstance>> eval
...
+        }
+        final JavaPairRDD<NullWritable, BytesWritable> rawRdd =
+                sparkExecutor.sc.sequenceFile(actualInputPath, NullWritable.class, BytesWritable.class);
+        final Class<T> typeClass = this.getTypeClass();
+        final JavaRDD<Object> rdd = rawRdd.flatMap(tuple -> {
+            byte[] payload = Arrays.copyOf(tuple._2.getBytes(), tuple._2.getLength());
+            try {
+                List<Object> chunk = ObjectFileSerialization.deserializeChunk(payload, serializationMode, typeClass);
```
juripetersen: Could be simplified to 

`return ObjectFileSerialization.deserializeChunk(payload, serializationMode, typeClass).iterator()`

wayang-platforms/wayang-spark/src/main/java/org/apache/wayang/spark/operators/SparkObjectFileSink.java
```diff
@@ -109,4 +125,34 @@ public boolean containsAction() {
         return true;
     }
 
+    private static Tuple2<NullWritable, BytesWritable> encodeBuffer(Object[] buffer,
+                                                                    int validLength,
+                                                                    ObjectFileSerializationMode mode) {
+        try {
+            byte[] payload = ObjectFileSerialization.serializeChunk(buffer, validLength, mode);
+            BytesWritable writable = new BytesWritable(payload);
+            return new Tuple2<>(NullWritable.get(), writable);
```
juripetersen: Could be simplified to: 

```java

return new Tuple2<>(NullWritable.get(), new BytesWritable(payload));
```

Definitely needs a blank line before the return.

wayang-api/wayang-api-python/pom.xml
```diff
@@ -37,6 +37,7 @@
         <spark.version>2.4.0</spark.version>
         <scala.mayor.version>2.12</scala.mayor.version>
         <giraph.version>1.2.0-hadoop2</giraph.version>
+        <python.worker.tests.skip>true</python.worker.tests.skip>
```
juripetersen: Are you adding this because the tests fail locally for you? 
I don't think this should be addressed in this PR, it has nothing to do with the python API

mspruc:
What about our usage of pickle.loads in:
https://github.com/apache/incubator-wayang/blob/d4fa09229e8c7f063212df509cea94dace84bb5e/python/src/pywy/execution/worker.py#L105 ?

wayang-api/wayang-api-python/pom.xml
novatechflow: It fails int he git workflow ... maybe you can fix that.

juripetersen: Our other recent PRs don't seem to have that problem. @mspruc, what is your take on this?

juripetersen:
One general question that remains for me:
@2pk03 What is the overhead needed for users to enable the JSON serialization, if any?

If there is none, why do we even keep offering the legacy serialization? Given that it poses as a potential security threat, it should be removed if there is no overhead.

wayang-api/wayang-api-python/pom.xml
mspruc: This PR has been opened with it in the initial commit, it is kind of hard for me to comment on it without having the stack trace.

novatechflow:
Tested with Apache Flink, there is no overhead from source and sink's side. It's just to use the serializer - physically, like computation power and memory, it adds the burden to Wayang's processing when filtering, but I think it's marginal tbh,

wayang-api/wayang-api-python/pom.xml
novatechflow: I'll remove and run again, maybe the pipeline had a issue.

mspruc:
> What about our usage of pickle.loads in:
> 
> https://github.com/apache/incubator-wayang/blob/d4fa09229e8c7f063212df509cea94dace84bb5e/python/src/pywy/execution/worker.py#L105
> ?

@2pk03 Do you know if this also needs to be changed?

novatechflow:
> > What about our usage of pickle.loads in:
> > https://github.com/apache/incubator-wayang/blob/d4fa09229e8c7f063212df509cea94dace84bb5e/python/src/pywy/execution/worker.py#L105
> > 
> > ?
> 
> @2pk03 Do you know if this also needs to be changed?

the pickle.loads(decoded_udf) mirrors how PySpark workers hydrate the UDF sent from the JVM. That payload comes from our own driver, right?

mspruc:
Yeah, this is from our own driver, I guess by reasoning it should be fine

juripetersen (approved):

PR #646, opened 2025-12-15
Feature/spark dataframes

novatechflow:
## Summary

- Add Spark Dataset/DataFrame plumbing: Parquet source/sink flag, channel conversions, optimizer cost hints.
- Document how to build dataset-backed pipelines (`README.md`, `guides/spark-datasets.md`).

## Next steps / follow-ups

- ML4All pipelines still emit/consume raw `double[]`/`Double` RDDs. We should extend them to use `DatasetChannel`s once schema handling is in place.
- Text/Object sources currently produce RDD channels. A Record-backed variant (or a conversion helper) would allow dataset output without extra user code.

### DIFF
--- a/wayang-api/wayang-api-scala-java/src/main/scala/org/apache/wayang/api/JavaPlanBuilder.scala
+++ b/wayang-api/wayang-api-scala-java/src/main/scala/org/apache/wayang/api/JavaPlanBuilder.scala
@@ -71,7 +71,7 @@ class JavaPlanBuilder(wayangCtx: WayangContext, jobName: String) {
    * @return [[DataQuantaBuilder]] for the file
    */
   def readParquet(url: String,
-                  projection: Array[String] = null): UnarySourceDataQuantaBuilder[UnarySourceDataQuantaBuilder[_, Record], Record] =
+                  projection: Array[String]): UnarySourceDataQuantaBuilder[UnarySourceDataQuantaBuilder[_, Record], Record] =
     readParquet(url, projection, preferDataset = false)
 
   def readParquet(url: String,


wayang-api/wayang-api-scala-java/src/main/scala/org/apache/wayang/api/DataQuanta.scala
```diff
@@ -1027,6 +1027,12 @@ class DataQuanta[Out: ClassTag](val operator: ElementaryOperator, outputIndex: I
     writeTextFileJava(url, toSerializableFunction(formatterUdf), udfLoad)
   }
 
+  def writeParquet(url: String, overwrite: Boolean = false)(implicit ev: Out =:= Record): Unit =
+    writeParquetJava(url, overwrite, preferDataset = false)
+
+  def writeParquetAsDataset(url: String, overwrite: Boolean = true)(implicit ev: Out =:= Record): Unit =
```
juripetersen: Is it nicer for the API here if `writeParquet` gets a `preferDataset` parameter that is false by default? 
This would allow to remove `writeParquetAsDataset` altogether.

wayang-api/wayang-api-scala-java/src/main/scala/org/apache/wayang/api/JavaPlanBuilder.scala
```diff
@@ -72,6 +72,16 @@ class JavaPlanBuilder(wayangCtx: WayangContext, jobName: String) {
...
     createSourceBuilder(ParquetSource.create(url, projection))(ClassTag(classOf[Record]))
 
+  /**
+    * Read a parquet file and provide it as a dataset of [[Record]]s backed by Spark Datasets.
+    *
+    * @param url the URL of the Parquet file
+    * @param projection the projection, if any
+    * @return [[DataQuantaBuilder]] for the file
+    */
+  def readParquetAsDataset(url: String, projection: Array[String] = null): UnarySourceDataQuantaBuilder[UnarySourceDataQuantaBuilder[_, Record], Record] =
```
juripetersen: The same thought goes for this. 

We could use `readParquet` with the flag here as well.

wayang-api/wayang-api-scala-java/src/main/scala/org/apache/wayang/api/PlanBuilder.scala
```diff
@@ -140,6 +140,16 @@ class PlanBuilder(private[api] val wayangContext: WayangContext, private var job
...
   def readParquet(url: String, projection: Array[String] = null): DataQuanta[Record] = load(ParquetSource.create(url, projection))
 
+  /**
+   * Read a parquet file and keep it backed by a Spark Dataset throughout execution.
+   *
+   * @param url the URL of the Parquet file
+   * @param projection the projection, if any
+   * @return [[DataQuanta]] of [[Record]]s backed by a Spark Dataset when executed on Spark
+   */
+  def readParquetAsDataset(url: String, projection: Array[String] = null): DataQuanta[Record] =
```
juripetersen: The same thought goes for this.

We could use `readParquet` with the flag here as well.

novatechflow: good catch - yeah, good idea. Its a bit duplicated given some month of coding :)

wayang-platforms/wayang-spark/src/main/java/org/apache/wayang/spark/channels/ChannelConversions.java
```diff
@@ -119,6 +150,24 @@ public class ChannelConversions {
...
             CACHED_RDD_TO_HDFS_TSV,
-            UNCACHED_RDD_TO_HDFS_TSV
+            UNCACHED_RDD_TO_HDFS_TSV,
+            DATASET_TO_UNCACHED_RDD,
+            CACHED_DATASET_TO_UNCACHED_RDD,
+            UNCACHED_RDD_TO_UNCACHED_DATASET,
+            CACHED_RDD_TO_UNCACHED_DATASET
     );
+
+    private static SparkRddToDatasetOperator createRddToDatasetOperator(Channel sourceChannel,
```
mspruc: Can you explain this method? I'm not sure I completely understand the motivation here

wayang-api/wayang-api-scala-java/src/main/scala/org/apache/wayang/api/JavaPlanBuilder.scala
```diff
@@ -69,18 +69,22 @@ class JavaPlanBuilder(wayangCtx: WayangContext, jobName: String) {
    * @param projection the projection, if any
    * @return [[DataQuantaBuilder]] for the file
    */
-  def readParquet(url: String, projection: Array[String] = null): UnarySourceDataQuantaBuilder[UnarySourceDataQuantaBuilder[_, Record], Record] =
-    createSourceBuilder(ParquetSource.create(url, projection))(ClassTag(classOf[Record]))
+  def readParquet(url: String,
```
juripetersen, line 77: Couldn't this entire method be removed if both projection and preferDataset had default arguments? There would be no need for overloading this, right?

wayang-api/wayang-api-scala-java/src/main/scala/org/apache/wayang/api/PlanBuilder.scala
```diff
@@ -136,19 +136,13 @@ class PlanBuilder(private[api] val wayangContext: WayangContext, private var job
...
-  /**
-   * Read a parquet file and keep it backed by a Spark Dataset throughout execution.
-   *
-   * @param url the URL of the Parquet file
-   * @param projection the projection, if any
-   * @return [[DataQuanta]] of [[Record]]s backed by a Spark Dataset when executed on Spark
-   */
-  def readParquetAsDataset(url: String, projection: Array[String] = null): DataQuanta[Record] =
-    load(ParquetSource.create(url, projection).preferDatasetOutput(true))
+  def readParquet(url: String,
```
juripetersen, line 142: Nice, this is more like I think it should be. Maybe you can just change it to this style in JavaPlanBuilder as well for a uniform style.

juripetersen (dismissed):

novatechflow:
Fixed the overload issue. Java does not see the Scale overload. Build run through now.

novatechflow:
This is a likely Scala version mismatch. `scala.annotation.JvmOverloads` isn’t available in the Scala version used by CI (likely 2.11), but my local build uses a newer Scala, so it compiled locally with Scala version 2.12.17 - it's a bit surprising since it runs locally. 

```mvn -q -Dexpression=scala.version help:evaluate -DforceStdout```
`2.12.17% `

```mvn -pl wayang-api/wayang-api-scala-java -am -DskipTests compile```
```[INFO] BUILD SUCCESS```

juripetersen (approved):

Issue #534, opened 2025-04-09
JSON API doesnt return an error mesaage


jephal:
when sending an incorrect Wayang plan to the JSON API endpoint it throws an error in terminal instead of returning error

alisha-1000:
Thanks for reporting this ,

It sounds like the JSON API currently logs the exception to the server terminal but does not propagate a structured error response back to the client when an invalid Wayang plan is submitted. This can make debugging difficult for API consumers.

Ideally, the endpoint should:

Catch plan parsing / validation exceptions

Return a proper HTTP status code (e.g., 400 Bad Request for invalid plans, 500 for internal errors)

Provide a structured JSON error response (e.g., { "error": "...", "details": "..." })

Avoid exposing internal stack traces unless running in debug mode

This would make the API more robust and production-ready.

If you’d like, I can look into adding proper exception handling around the endpoint and implementing consistent error responses.

zkaoudi:
thanks a lot @alisha-1000. This has been addressed recently but without closing the issue (check PR #651). Thank you for catching that!

PR #651, opened 2025-12-30
Modified REST server in JSON-REST API

aleudework:
Changed the REST server (Main.scala) in JSON-REST API to also return the error message back to client when an execution error occur. Only status code are returned originally.

This is for agentic setups to be able to debug generated JSON Wayang plans.


zkaoudi:
this closes issue #534

PR #652, opened 2026-01-06
Fix JOIN clause to reference the correct table in SQL query generation

treyfel:
* Corrected the SQL JOIN clause in `JdbcJoinOperator#createSqlClause` to use the right table name when forming the JOIN statement, ensuring the SQL query generated is not a self-join.

* Updated the expected SQL query in `JdbcJoinOperatorTest#testWithHsqldb` to match the corrected JOIN clause logic.

### DIFF
--- a/wayang-platforms/wayang-jdbc-template/src/main/java/org/apache/wayang/jdbc/operators/JdbcJoinOperator.java
+++ b/wayang-platforms/wayang-jdbc-template/src/main/java/org/apache/wayang/jdbc/operators/JdbcJoinOperator.java
@@ -73,7 +73,7 @@ public abstract class JdbcJoinOperator<KeyType>
         final String rightTableName = right.field0;
         final String rightKey = right.field1;
 
-        return "JOIN " + leftTableName + " ON " +
+        return "JOIN " + rightTableName + " ON " +
             rightTableName + "." + rightKey
             + "=" + leftTableName + "." + leftKey;
     }
--- a/wayang-platforms/wayang-jdbc-template/src/test/java/org/apache/wayang/jdbc/operators/JdbcJoinOperatorTest.java
+++ b/wayang-platforms/wayang-jdbc-template/src/test/java/org/apache/wayang/jdbc/operators/JdbcJoinOperatorTest.java
@@ -119,7 +119,7 @@ class JdbcJoinOperatorTest extends OperatorTestBase {
         System.out.println();
 
         assertEquals(
-            "SELECT * FROM testA JOIN testA ON testB.a=testA.a;",
+            "SELECT * FROM testA JOIN testB ON testB.a=testA.a;",
             sqlQueryChannelInstance.getSqlQuery()
         );
     }


mspruc:
ty for your contribution,

I'm not entirely sure I understand the issue you are trying to fix here, `JdbcJoinOperator.createSqlClause()` now uses the right table name twice, but what about the left table name then? 

Can I ask what issue prompted the change? If you have a stack trace or something I can also take a look at it.

wayang-platforms/wayang-jdbc-template/src/main/java/org/apache/wayang/jdbc/operators/JdbcJoinOperator.java
kamir, line 1: I think the problem was in the test class:

SELECT * FROM **testA** JOIN **testA** ON testB.a=testA.a;

In [JdbcJoinOperator.java‎](https://github.com/apache/wayang/pull/652/changes#diff-daca6094ef619463d8bbaa8ec264dc4723b92352f4646de638478f642c7cbc47) 
the code looks good.

wayang-platforms/wayang-jdbc-template/src/main/java/org/apache/wayang/jdbc/operators/JdbcJoinOperator.java
```diff
@@ -73,7 +73,7 @@ public String createSqlClause(Connection connection, FunctionCompiler compiler)
         final String rightTableName = right.field0;
         final String rightKey = right.field1;
 
-        return "JOIN " + leftTableName + " ON " +
+        return "JOIN " + rightTableName + " ON " +
```
kamir, line 76:
```suggestion
        return "JOIN " + leftTableName + " ON " +
```
I think this is wrong and not needed. Just fix the test case where the assertion has teh same tablename twice.

treyfel: Hi! So previously the sql clause created would be a self join, which I assume was not intended. In `JdbcExecutor.createSqlString` the sql string is built. In line 217 in `JdbcExecutor`, a clause such as _"SELECT * FROM testA"_ is created. 

Then, _"JOIN testA ON testB.a=testA.a;"‎_ in line 221 using the `JdbcJoinOperator.createSqlClause()` is added. The change intends to address this issue by making the `JdbcJoinOperator.createSqlClause()` not reference the same table as in the selection  but rather the other table.

treyfel:
> ty for your contribution,
> 
> I'm not entirely sure I understand the issue you are trying to fix here, `JdbcJoinOperator.createSqlClause()` now uses the right table name twice, but what about the left table name then?
> 
The left table name comes from the tableSource `JdbcTableSource.createSqlClause()`.
> Can I ask what issue prompted the change? If you have a stack trace or something I can also take a look at it.
>
There was no initial issue as the tests we’re green but when working on something regarding join operations on the jdbc platform, I noticed the test checks for a self join syntax, which I assume was not intended.

zkaoudi:
+1 for this PR. 
The left table is created as a "FROM $leftTableName" statement in the source operator. In the proposed change the "JOIN" should be done on the right table as a string of "JOIN $rightTableName". Then the equality "ON" comes where it doesn't matter the order of the predicates.

@treyfel maybe you can split up the return string into two strings  (join part and predicates part) so that it's not confusing with a single return statement

mspruc:
let's merge this, I still have some outstanding issues with how we handle `createSqlClause()` and `.withSqlImplementation()` as ive previously mentioned in #643 but I have something for that in the pipeline, for now this should be fine @zkaoudi @kamir

zkaoudi:
Hi @kamir, can you merge the PR? The suggested changes are correct

kamir (approved):

PR #657, opened 2026-01-15
Move functional interfaces & deprecate createSql

mspruc:


juripetersen:
@mspruc you need to add license headers to your files.

wayang-commons/wayang-core/src/main/java/org/apache/wayang/core/function/SerializableFunctionInterface.java
```diff
@@ -0,0 +1,98 @@
+package org.apache.wayang.core.function;
```
juripetersen, line 19: Add license header.

juripetersen (approved):

juripetersen:
@mspruc feel free to merge this if you think it should be.

mspruc:
since it's a structural change ill wait till it gets another approval @juripetersen

Issue #594, opened 2025-06-25
Non-deterministic test failure in TensorflowIrisIT


mspruc:
[TensorFlowIrisIT test](https://github.com/apache/incubator-wayang/blob/6db324b973314efc940dab1f2b388b40d4d8bed8/wayang-tests-integration/src/test/java/org/apache/wayang/tests/TensorflowIrisIT.java#L60) can non-deterministically fail with error:

```
org.apache.wayang.core.api.exception.WayangException: Executing T[JavaCollect[convert out@JavaMap[1->1, id=400f7211]]] failed.
        at org.apache.wayang.java.execution.JavaExecutor.execute(JavaExecutor.java:92)
        at org.apache.wayang.core.platform.PushExecutorTemplate.execute(PushExecutorTemplate.java:73)
        at org.apache.wayang.core.platform.PushExecutorTemplate$StageExecution.execute(PushExecutorTemplate.java:195)
        at org.apache.wayang.core.platform.PushExecutorTemplate$StageExecution.doExecute(PushExecutorTemplate.java:166)
        at org.apache.wayang.core.util.OneTimeExecutable.tryExecute(OneTimeExecutable.java:41)
        at org.apache.wayang.core.util.OneTimeExecutable.execute(OneTimeExecutable.java:54)
        at org.apache.wayang.core.platform.PushExecutorTemplate$StageExecution.executeStage(PushExecutorTemplate.java:156)
        at org.apache.wayang.core.platform.PushExecutorTemplate.execute(PushExecutorTemplate.java:61)
        at org.apache.wayang.core.platform.CrossPlatformExecutor.execute(CrossPlatformExecutor.java:378)
        at org.apache.wayang.core.platform.CrossPlatformExecutor.executeSingleStage(CrossPlatformExecutor.java:248)
        at org.apache.wayang.core.platform.CrossPlatformExecutor.runToBreakpoint(CrossPlatformExecutor.java:320)
        at org.apache.wayang.core.platform.CrossPlatformExecutor.executeUntilBreakpoint(CrossPlatformExecutor.java:156)
        at org.apache.wayang.core.api.Job.execute(Job.java:542)
        at org.apache.wayang.core.api.Job.doExecute(Job.java:326)
        at org.apache.wayang.core.util.OneTimeExecutable.tryExecute(OneTimeExecutable.java:41)
        at org.apache.wayang.core.util.OneTimeExecutable.execute(OneTimeExecutable.java:54)
[...]

PR #658, opened 2026-01-15
codeql version

mspruc:
old version was deprecated and is now causing build errors


mspruc:
@juripetersen rerun the build, fails on #594

PR #661, opened 2026-01-19
Source, map, collection, sink for bounded data streams in Flink

### LINKED ISSUE #615: Port Flink-DataSet to Flink-DataStream as DataSet is deprecated
Flink provides a migration guide:

https://nightlies.apache.org/flink/flink-docs-release-1.20/docs/dev/datastream/dataset_migration/

It is not trivial to port every operator, but the standard operations like map, filter, reduce and join should be easy to port. 
Providing this as a seperate platform called `flink-stream` might be a way to ease development as you could avoid rewriting the `FlinkExecutor` and "just" write a new one from scratch that is built with `DataStream` in mind.

mspruc:
lets have a discussion about how we are going to handle Continuous sources e.g. in flink. And how we are going to integrate this vs the old data set implementation.

resolves #615


novatechflow:
yes, pls start in our dev mailing list.

mspruc:
Please LMK if I need to deprecate any parts of the old DataSet API.

also I'm unsure whether its a smart idea to implement `Join` for both continuous and bounded streams like this, thoughts?

mspruc:
@juripetersen  @zkaoudi  @novatechflow Review when you have time pls

wayang-platforms/wayang-flink/src/main/java/org/apache/wayang/flink/channels/DataStreamChannel.java
```diff
@@ -0,0 +1,103 @@
...
+
+    /**
+     * {@link ChannelInstance} implementation for {@link DataStream}s.
+     */
+    public class Instance extends AbstractChannelInstance {
+
+        private DataStream<?> dataStream;
+
+        // TODO: this.size is currently always 0
+        private long size;
```
juripetersen, line 43: Does this have any effects?

wayang-platforms/wayang-flink/src/main/java/org/apache/wayang/flink/execution/FlinkExecutor.java
```diff
@@ -56,6 +56,12 @@ public class FlinkExecutor extends PushExecutorTemplate {
      */
     public ExecutionEnvironment fee;
```
juripetersen, line 57: I think ideally we just remove the old Env and also DataSet with it, so that the end user of wayang doesn't even notice this change (other than a few operators missing maybe)

wayang-platforms/wayang-flink/src/main/java/org/apache/wayang/flink/mapping/BoundedTextFileSourceMapping.java
```diff
@@ -0,0 +1,55 @@
...
+import org.apache.wayang.basic.operators.TextFileSource;
+import org.apache.wayang.core.mapping.Mapping;
+import org.apache.wayang.core.mapping.OperatorPattern;
+import org.apache.wayang.core.mapping.PlanTransformation;
+import org.apache.wayang.core.mapping.ReplacementSubplanFactory;
+import org.apache.wayang.core.mapping.SubplanPattern;
+import org.apache.wayang.flink.operators.FlinkBoundedTextFileSource;
+import org.apache.wayang.flink.platform.FlinkPlatform;
+
+public class BoundedTextFileSourceMapping implements Mapping {
```
juripetersen, line 33: As everything is bounded as of now in Wayang, I would just call this TextFileSource and replace TextFileSource from Flink with it. 

We can later add a StreamedTextFileSource.

wayang-platforms/wayang-flink/src/main/java/org/apache/wayang/flink/mapping/StreamedJoinMapping.java
```diff
@@ -0,0 +1,56 @@
...
+import org.apache.wayang.core.mapping.ReplacementSubplanFactory;
+import org.apache.wayang.core.mapping.SubplanPattern;
+import org.apache.wayang.core.types.DataSetType;
+import org.apache.wayang.flink.operators.FlinkDataStreamJoinOperator;
+import org.apache.wayang.flink.platform.FlinkPlatform;
+
+/**
+ * Mapping from {@link JoinOperator} to {@link FlinkDataStreamJoinOperator}.
+ */
+public class StreamedJoinMapping implements Mapping {
```
juripetersen, line 37: Same goes for this operator. I think we should just remove the DataSet ones and then replace them with the DataStream.

juripetersen (changes requested):
I left some comments, but we should just discuss if that's the way to go.

wayang-platforms/wayang-flink/src/main/java/org/apache/wayang/flink/channels/DataStreamChannel.java
mspruc: Leftovers from the old `DataSetChannel` implementation the user could theoretically extend the DataStreamChannel and provide their own, but its such a niche situation I'm fine with removing it.

wayang-platforms/wayang-flink/src/main/java/org/apache/wayang/flink/execution/FlinkExecutor.java
mspruc: Per discussion in the dev-list I've delegated this to the configuration.

wayang-platforms/wayang-flink/src/main/java/org/apache/wayang/flink/mapping/BoundedTextFileSourceMapping.java
mspruc: See comment above.

wayang-platforms/wayang-flink/src/main/java/org/apache/wayang/flink/mapping/StreamedJoinMapping.java
mspruc: See comment above.

wayang-platforms/wayang-flink/src/main/java/org/apache/wayang/flink/mapping/BoundedTextFileSourceMapping.java
mspruc: Nevertheless since our discussion on continuous streams hasn't concluded either I still propose we keep the bounded semantic as a way to leave the door open for continuous sources later.

juripetersen:
I guess we just need more activity in the discussion on the mailing list.

juripetersen (approved):

PR #647, opened 2025-12-16
#634

novatechflow:
this should fix the non-deterministic enumeration. Main issue from my point of view was that we relied completely on plain `HashSet` all over the place => the order changed every time the JVM ran, therefore a unpredictable hash-set ordering was the root cause.

Wrote a regression test, too:
`mvn -pl wayang-commons/wayang-core -Dtest=PlanEnumerationDeterminismTest,ChannelConversionGraphDeterminismTest test`

Issue #634, opened 2025-11-18
Enumeration is non-deterministic


juripetersen:
Using Wayangs default enumeration and pruning strategies, the plans compared in cost models are non-deterministic. 
This can be observed when comparing costs with different mechanisms:

```java
            final PlanImplementation bestPlanImplementation = executionPlans.stream()
                .reduce((p1, p2) -> {
                    final double t1 = p1.getSquashedCostEstimate();
                    final double t2 = p2.getSquashedCostEstimate();
                    return t1 < t2 ? p1 : p2;
                })

```

novatechflow:
Ya stumbled upon as I tested #646 - think about to replace the HashSet usage in PlanEnumeration/MultiMap with LinkedHashSet, or convert the sets into deterministically sorted Lists (e.g., order by operator IDs) before feeding to the collector.

juripetersen:
We already did exactly this in a fork of ours. However, that didn't solve the non-deterministic nature.

novatechflow:
Yeah, there are a lot of unordered collections across the channels, I saw the helper keeps rehashing, that's the main issue I think. Side effects then unordered collections, which let the graph go havoc. I write a regression test today to find other places.

mspruc:
with #662 is this issue resolved @juripetersen @novatechflow ?

mspruc:
closing this for now, reopen if the issue persists

PR #662, opened 2026-01-19
Ensure deterministic plan enumeration and channel conversion

novatechflow:
Separates determinism fixes/tests from Spark DataFrames work. Cherry-picks commits 7bcf2ff and 93fbadb from PR #647.
- Stable comparator + ordered sets for deterministic selection
- Adds determinism tests for plan enumeration and channel conversion

solves #634

### DIFF
--- a/wayang-commons/wayang-core/src/main/java/org/apache/wayang/core/optimizer/enumeration/PlanImplementation.java
+++ b/wayang-commons/wayang-core/src/main/java/org/apache/wayang/core/optimizer/enumeration/PlanImplementation.java
@@ -51,6 +51,7 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedHashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
@@ -253,7 +254,7 @@ public class PlanImplementation {
 
             // Discern LoopHeadOperator InputSlots and loop body InputSlots.
             final List<LoopImplementation.IterationImplementation> iterationImpls = loopImplementation.getIterationImplementations();
-            final Collection<InputSlot<?>> collector = new HashSet<>(innerInputs.size());
+            final Collection<InputSlot<?>> collector = new LinkedHashSet<>(innerInputs.size());
             for (InputSlot<?> innerInput : innerInputs) {
                 if (innerInput.getOwner() == loopSubplan.getLoopHead()) {
                     final LoopImplementation.IterationImplementation initialIterationImpl = iterationImpls.get(0);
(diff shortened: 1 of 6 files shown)


wayang-commons/wayang-core/src/main/java/org/apache/wayang/core/optimizer/enumeration/PlanImplementation.java
```diff
@@ -976,6 +992,67 @@ Stream<ExecutionOperator> streamOperators() {
...
+        final String descriptor = describeOperator(loop);
+        final String iterationDescriptor = implementation.getIterationImplementations().stream()
+                .map(iteration -> Integer.toString(iteration.getNumIterations()))
+                .collect(Collectors.joining(","));
+        return descriptor + ":" + iterationDescriptor;
+    }
+
+    private static String describeInputSlot(InputSlot<?> slot) {
+        if (slot == null) {
+            return "null";
```
juripetersen, line 1042: Couldn't this just return an actual null then? String is nullable, so this can be a valid value.
Or would this be a problem on comparison later on?

wayang-commons/wayang-core/src/test/java/org/apache/wayang/core/optimizer/channels/ChannelConversionGraphDeterminismTest.java
```diff
@@ -0,0 +1,137 @@
...
+        descriptorList.add(describeChannel(junction.getSourceChannel(), true));
+        for (int i = 0; i < junction.getNumTargets(); i++) {
+            descriptorList.add(describeChannel(junction.getTargetChannel(i), false));
+        }
+        return descriptorList;
+    }
+
+    private static String describeChannel(Channel channel, boolean isSourceChannel) {
+        if (channel == null) {
+            return "null";
```
juripetersen, line 114: Same as the comment about null on other describeX above.

wayang-commons/wayang-core/src/main/java/org/apache/wayang/core/optimizer/enumeration/PlanImplementation.java
novatechflow: Returning a "real" `null` would break the deterministic identifier: the values are `.sorted()` and then joined, and `sorted()` throws on nulls. The literal `"null"` keeps it stable and safe for comparison/concatenation. We could switch to real nulls only if we add null-safe sorting/joining.

wayang-commons/wayang-core/src/test/java/org/apache/wayang/core/optimizer/channels/ChannelConversionGraphDeterminismTest.java
novatechflow: https://github.com/apache/wayang/pull/662#discussion_r2711765367

juripetersen (approved):

PR #688, opened 2026-02-13
Fix async worker classpath/module args and fail on child error. fixes #687

### LINKED ISSUE #687: Bug: `multiDataQuantaExecuteTest` writes empty output on newer macOS
## Summary
`OtherSerializationTests.multiDataQuantaExecuteTest` intermittently fails on newer macOS with:

- expected lines: `3`
- actual lines: `0`

Linux and older macOS pass.

## Failure
- Module: `wayang-api/wayang-api-scala-java`
- Test: `org.apache.wayang.api.serialization.OtherSerializationTests#multiDataQuantaExecuteTest`
- Error: `Number of lines in the file should match ==> expected: <3> but was: <0>`

## Root Cause
Async execution spawns a child JVM in `org.apache.wayang.api.async.runAsyncBody` but:
1. Uses fragile classpath resolution (`java.class.path`) under Surefire isolation.
2. Does not forward parent JVM module access args (`--add-opens`, `--add-exports`).
3. Does not fail when child exits non-zero.

When the child fails to start/execute correctly, parent flow still continues, and output files remain empty.

## Expected
If async worker fails, test should fail with explicit worker error (not empty-file assertion), and classpath/module args should be propagated so worker can run consistently across OS/JDK combinations.

## Proposed Fix
- Build child classpath from context `URLClassLoader` (fallback to `java.class.path`).
- Forward parent `--add-opens` / `--add-exports` args.
- Throw on non-zero child exit code.

## Changed File
- `wayang-api/wayang-api-scala-java/src/main/scala/org/apache/wayang/api/async/package.scala`

novatechflow:
```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  04:54 min
[INFO] Finished at: 2026-02-13T08:48:32+01:00
[INFO] ------------------------------------------------------------------------
```

```
uname -smv
Darwin Darwin Kernel Version 25.3.0: Wed Jan 28 20:53:01 PST 2026; root:xnu-12377.81.4~5/RELEASE_ARM64_T8103 arm64
````

### DIFF
--- a/wayang-api/wayang-api-scala-java/src/main/scala/org/apache/wayang/api/async/package.scala
+++ b/wayang-api/wayang-api-scala-java/src/main/scala/org/apache/wayang/api/async/package.scala
@@ -22,8 +22,12 @@ import org.apache.logging.log4j.{LogManager, Logger}
 import org.apache.wayang.api.serialization.TempFileUtils
 import org.apache.wayang.core.api.exception.WayangException
 
+import java.io.File
+import java.lang.management.ManagementFactory
+import java.net.URLClassLoader
 import java.nio.charset.StandardCharsets
 import java.nio.file.{Files, Paths}
+import scala.collection.JavaConverters._
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
 import scala.reflect.ClassTag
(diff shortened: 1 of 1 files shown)


juripetersen (approved):

mspruc:
@luckyasser Can you also review and see if it works for you?

luckyasser:
works for me now

PR #695, opened 2026-02-21
fix: RecordType.hashCode() inconsistent with equals() (#694)

### LINKED ISSUE #694: bug: RecordType.hashCode() violates equals/hashCode contract
Hi team,

I was going through the code to get familiar and found a bug in `RecordType.hashCode()`.

The `equals()` method uses `Arrays.equals()` to compare field names by content but `hashCode()` passes the `fieldNames` array directly to `Objects.hash()` which ends up using the array's identity hash instead of its content. So basically two RecordType instances with the same field names are considered equal but will give different hashcodes which is wrong.

I wrote a small temp test locally as a sanity check:

    RecordType rt1 = new RecordType("x", "y", "z");
    RecordType rt2 = new RecordType("x", "y", "z");

    rt1.equals(rt2)  //true
    rt1.hashCode() == rt2.hashCode()  //false

    HashMap<RecordType, String> map = new HashMap<>();
    map.put(rt1, "z");
    map.get(rt2); //this will return null instead of "z"

The fix is pretty simple: line 68 in `RecordType.java` just needs to wrap the array.

mattpapp:
Fixes #694

`hashCode()` was using identity hash for the `fieldNames` array instead of content hash, while `equals()` correctly uses `Arrays.equals()`. Wrapped it with `Arrays.hashCode()` and added a test.

### DIFF
--- a/wayang-commons/wayang-basic/src/main/java/org/apache/wayang/basic/types/RecordType.java
+++ b/wayang-commons/wayang-basic/src/main/java/org/apache/wayang/basic/types/RecordType.java
@@ -65,7 +65,7 @@ public class RecordType extends BasicDataUnitType<Record> {
 
     @Override
     public int hashCode() {
-        return Objects.hash(super.hashCode(), fieldNames);
+        return Objects.hash(super.hashCode(), Arrays.hashCode(fieldNames));
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


mspruc:
Hiya thanks for your contribution,

Do you know if `BasicDataUnitType` also needs to specify `super.hashCode()` in its implementation of `hashCode()`?

mattpapp:
@mspruc Hey! good question.

I don’t think `BasicDataUnitType` should include `super.hashCode()`. Its superclass (`DataUnitType`) doesn’t override `hashCode`/`equals` (adn has no state) so `super.hashCode()` would basically be `Object.hashCode()`  and would make two equal `BasicDataUnitType` instances give different hashes.

`RecordType` including `super.hashCode()` is still correct since `RecordType.equals()` calls `super.equals()` (so the parent state/typeClass is part of equality)

PR #696, opened 2026-02-21
Move JavaCSVTableSource to Java platform and rename to JavaCSVFileSource

### LINKED ISSUE #482: Move JavaCSVTableSource operator from sql-api to the Java platform
There is a Java csv source operator inside the wayang sql api:
[JavaCSVTableSource](https://github.com/apache/incubator-wayang/blob/main/wayang-api/wayang-api-sql/src/main/java/org/apache/wayang/api/sql/sources/fs/JavaCSVTableSource.java).

We should move it to the wayang-platforms/wayang-java module, specifically [here](https://github.com/apache/incubator-wayang/tree/main/wayang-platforms/wayang-java/src/main/java/org/apache/wayang/java/operators) and rename it to JavaCSVFileSource because it is reading from a file and not a table in the database.

Chandanakt:
This PR moves JavaCSVTableSource from the SQL API module to the
wayang-java platform module and renames it to JavaCSVFileSource.

The helper class CsvRowConverter was also moved accordingly.
No functional changes were introduced.

Fixes #482

### DIFF
--- a/wayang-api/wayang-api-sql/src/main/java/org/apache/wayang/api/sql/calcite/converter/WayangTableScanVisitor.java
+++ b/wayang-api/wayang-api-sql/src/main/java/org/apache/wayang/api/sql/calcite/converter/WayangTableScanVisitor.java
@@ -23,7 +23,7 @@ import org.apache.calcite.rel.type.RelDataTypeField;
 
 import org.apache.wayang.api.sql.calcite.rel.WayangTableScan;
 import org.apache.wayang.api.sql.calcite.utils.ModelParser;
-import org.apache.wayang.api.sql.sources.fs.JavaCSVTableSource;
+import org.apache.wayang.java.operators.JavaCSVFileSource;
 import org.apache.wayang.core.plan.wayangplan.Operator;
 import org.apache.wayang.core.types.DataSetType;
 import org.apache.wayang.jdbc.operators.JdbcTableSource;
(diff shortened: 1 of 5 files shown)


zkaoudi:
Hi @Chandanakt, there were some updates in the specified file, can you pull the changes and resolve conflicts?

As I can see from your PR we have some dependency on Calcite which we would have to add in the wayang-java module. Do you think there is a way to avoid that? It would be preferable if Calcite is used only in the SQL module

Chandanakt:
Hi @zkaoudi , thanks for the feedback.
I will rebase my branch on the latest main and resolve any conflicts.

Regarding the Calcite dependency: I agree with your suggestion. I'll refactor the implementation so that the SQL module maps RelDataType to a lightweight internal representation, and JavaCSVFileSource no longer depends on Calcite and will operate only on that representation. This keeps Calcite usage confined to the SQL module while preserving the existing conversion logic.

I’ll work on these changes and update the PR shortly. Thanks!

Chandanakt:
Hi @zkaoudi ,
I’ve pushed the updated changes.
The SQL module now maps RelDataType to a lightweight CsvType, and JavaCSVFileSource no longer depends on Calcite.
Please let me know if any further adjustments are needed. Thanks!

mspruc:
Hi, can you resolve the merge conflicts?

PR #699, opened 2026-02-26
Extend Wayang with spatial operators

MaxSpeer:
Hi everyone,
together with @zkaoudi we worked on extending wayang with spatial operations. 
Please consider our PR. We are looking forward to your feedback. 

## Summary
This PR introduces spatial processing capabilities in Wayang by adding two new logical operators: SpatialFilter and SpatialJoin, with platform-specific implementations for Java, Spark, and PostgreSQL/JDBC.
The goal is to enable spatial filter and join operations end-to-end across Wayang’s Java, Spark, and Postgres execution paths and remain extendable for other spatial operations.

## What Changed
- Added new logical operators in the common layer: SpatialFilterOperator and SpatialJoinOperator.
- Added spatial abstractions in core: SpatialPredicate and SpatialGeometry.
- Added a dedicated spatial plugin module: wayang-plugins/wayang-spatial.
- Added mappings and operators for: Java, Spark, and PostgreSQL/JDBC.
- Extended Java API with: spatialFilter(...) and spatialJoin(...).
- Added GeoJsonFileSource support in the spatial plugin (logical + Java implementation).
- Extended JDBC SQL planning/execution path to include spatial filter/join operators in SQL query construction.

## Design Choices
1. Core remains only defines spatial contracts (SpatialGeometry, SpatialPredicate) and has no direct JTS/Sedona/PostGIS dependency.
2. Backend-specific dependencies are isolated in the spatial plugin JTS/Sedona/PostGIS specifics are implemented in wayang-spatial, keeping the API/core clean and modular.

## Platform Implementation Notes
- Java: Uses JTS-based predicate evaluation and spatial index support for joins.
- Spark: Uses Apache Sedona for distributed spatial filtering and joining.
- PostgreSQL/JDBC: Generates spatial SQL clauses, enabling DB-side execution of filter/join. (Limitation: JDBCExecutor does not support multiple sources at the moment)

Testing
- Unit tests for: WayangGeometry, Java operators, Spark operators, JDBC SQL generation.
- Java API tests for: spatial filter/join usage and chained operations.
- Postgres integration tests are present but disabled by default due to external DB dependency.


mspruc:
Hi thanks for your contribution,
I will have a deeper look at your PR later but in the meantime, can you clean up the various `System.out.println(...)` statements?

MaxSpeer:
> Hi thanks for your contribution, I will have a deeper look at your PR later but in the meantime, can you clean up the various `System.out.println(...)` statements?

Hi @mspruc, the print statements we introduced are only in benchmark classes and in tests. We think this is fine. Could you please point to the statements that should be removed?

wayang-commons/wayang-basic/src/main/java/org/apache/wayang/basic/operators/SpatialJoinOperator.java
```diff
@@ -0,0 +1,109 @@
...
+import org.apache.wayang.core.api.spatial.SpatialGeometry;
+import org.apache.wayang.core.api.spatial.SpatialPredicate;
+import org.apache.wayang.core.function.FunctionDescriptor;
+import org.apache.wayang.core.function.TransformationDescriptor;
+import org.apache.wayang.core.optimizer.OptimizationContext;
+import org.apache.wayang.core.optimizer.cardinality.CardinalityEstimate;
+import org.apache.wayang.core.plan.wayangplan.BinaryToUnaryOperator;
+import org.apache.wayang.core.types.DataSetType;
+
+public class SpatialJoinOperator<InputType0, InputType1> extends BinaryToUnaryOperator<InputType0, InputType1, Tuple2<InputType0, InputType1>> {
```
zkaoudi, line 33: Add a comment about the operator similar to what you had done to spatialfilter operator

wayang-platforms/wayang-jdbc-template/src/main/java/org/apache/wayang/jdbc/execution/JdbcExecutor.java
```diff
@@ -170,21 +172,22 @@ protected static Tuple2<String, SqlQueryChannel.Instance> createSqlQuery(final E
...
         final Set<ExecutionTask> allTasks = stage.getAllTasks();
         assert allTasks.size() <= 3;
         ExecutionTask nextTask = JdbcExecutor.findJdbcExecutionOperatorTaskInStage(startTask, stage);
         while (nextTask != null) {
             // Evaluate the nextTask.
-            if (nextTask.getOperator() instanceof final JdbcFilterOperator filterOperator) {
-                filterTasks.add(filterOperator);
-            } else if (nextTask.getOperator() instanceof JdbcProjectionOperator projectionOperator) {
+            final var operator = nextTask.getOperator();
+            if (operator instanceof JdbcFilterOperator || operator instanceof SpatialFilterOperator) {
```
zkaoudi: Shouldn't this be a jdbcspatialfilter operator? Why referencing the logical operator?

zkaoudi: And maybe then we do not need the extra condition in the if?

wayang-platforms/wayang-jdbc-template/src/main/java/org/apache/wayang/jdbc/execution/JdbcExecutor.java
```diff
@@ -170,21 +172,22 @@ protected static Tuple2<String, SqlQueryChannel.Instance> createSqlQuery(final E
...
+                filterTasks.add((JdbcExecutionOperator) operator);
+            } else if (operator instanceof JdbcProjectionOperator) {
                 assert projectionTask == null; // Allow one projection operator per stage for now.
-                projectionTask = projectionOperator;
-            } else if (nextTask.getOperator() instanceof JdbcJoinOperator joinOperator) {
-                joinTasks.add(joinOperator);
+                projectionTask = (JdbcProjectionOperator) operator;
+            } else if (operator instanceof JdbcJoinOperator || (operator instanceof SpatialJoinOperator)) {
```
zkaoudi: same here? why a special case for the spatial wayang join operator is needed?

wayang-plugins/wayang-spatial/src/test/java/org/apache/wayang/spatial/integration/PostgresSpatialIntegrationTest.java
```diff
@@ -0,0 +1,284 @@
...
+import org.locationtech.jts.geom.GeometryFactory;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+@Disabled("Requires local Postgres test database.")
```
zkaoudi, line 47: can you check if the hsqldb library we use for other jdbc tests supports spatial operators? Then we can have the tests without requiring a postgres instance setup and running.

zkaoudi (changes requested):

zkaoudi:
And regarding the print outs, I agree with @mspruc. You can either remove them or you can homogenize them as some benchmarks print on the top and some others printout the count. But that's more of a minor.

wayang-platforms/wayang-jdbc-template/src/main/java/org/apache/wayang/jdbc/execution/JdbcExecutor.java
MaxSpeer: The JdbcSpatialOperator is located in the Spatial Plugin and we don't want to introduce this dependency. The findJdbcExecutionOperatorTaskInStage() function ensures that only Jdbc Operators are selected. For consistency we could change the other JDBC operators to their wayang operators (e.g. JdbcFilterOperator -> FilterOperator)

instant-sky:
We updated the spatial benchmarks with consistent print statements across the different classes. The benchmark scripts serve as examples on how to use the spatial operations and we use them to compare operator performance on different platforms.

wayang-plugins/wayang-spatial/src/test/java/org/apache/wayang/spatial/integration/PostgresSpatialIntegrationTest.java
MaxSpeer: hsqldb does not support spatial operations. In JdbcSpatialFilterOperatorTest we do however test if the generated Spatial SQL Query is correct.

wayang-platforms/wayang-jdbc-template/src/main/java/org/apache/wayang/jdbc/execution/JdbcExecutor.java
zkaoudi: Yes, I think having the Wayang operators is a good idea. In any case, we wouldn't have an execution operator from another platform here, so it's more about the type of operator and not about its execution platform. 

So, if we say eg. FilterOperator, would that also catch the case for the SpatialFilterOperator? It should, right?

zkaoudi (approved):

zkaoudi:
Thank you @treyfel @instant-sky @MaxSpeer for your contribution! 

@mspruc I'm good with this PR, if you are also fine, we can merge :)

wayang-plugins/wayang-spatial/src/main/java/org/apache/wayang/spatial/operators/jdbc/JdbcSpatialFilterOperator.java
```diff
@@ -0,0 +1,78 @@
...
+            throw new IllegalStateException("Geometry for spatial filter must not be null.");
+        }
+
+        // Column expression (e.g. "geom" or "t.geom")
+        final String columnExpr = this.keyDescriptor.getSqlImplementation().getField1();
+
+        // Geometry literal as ST_GeomFromText('WKT', srid)
+        final String wkt = this.referenceGeometry.toWKT();
+        // TODO: Check which SRID to use.
+        final int srid = 4326;
```
mspruc, line 66: Is this TODO resolved, or meant to be part of the PR?

mspruc (approved):
I think this is good for now, I have some reservations about the type system and allowing ```_: < SpatialGeometry``` but nothing we can't fix or support later on, good work.

wayang-plugins/wayang-spatial/src/main/java/org/apache/wayang/spatial/operators/jdbc/JdbcSpatialFilterOperator.java
mspruc: you can merge when this is resolved :)

treyfel: > Is this TODO resolved, or meant to be part of the PR?

Sorry for the late reply. This is yet to be implemented so I would argue to leave it in the PR.

zkaoudi: What is this srid meant to do? maybe a better explanation in the todo could help someone else to fix it in the future.

treyfel: Afaik it’s the id of the coordinate reference system with the value we have assigned being the id of the earths coordinate reference system. I agree that this could be made clearer in the comment.

PR #710, opened 2026-03-02
Support multiple field projection, by names and indexes for Pojo (#649)

JunWang222:
issue#649 Revise implementation to support multiple field projection, by names

- Add MultiFieldPojoImplementation static class following the existing PojoImplementation and RecordImplementation pattern
- Add createForPojoByNames() factory for projecting multiple POJO fields by name into a Record
- Add createForPojoByIndexes() factory for projecting POJO fields by their declared index into a Record
- Restore RecordImplementation and createRecordJavaImplementation for consistency with the original codebase style
- Add 9 new tests covering multi-field by name, by index, field reordering, null values, out-of-bounds, and nonexistent field error cases

Plan
Add a new MultiFieldPojoImplementation class following the same pattern as the existing PojoImplementation and RecordImplementation. It takes multiple field names, extracts each field from the input POJO via reflection, and packs the values into a Record.
Expose it through two new factory methods — one for field names, one for indexes — and keep the existing single-field path untouched for backward compatibility. The existing constructor routes automatically: 
* one field → old PojoImplementation (returns the raw value, same behavior as before)
* multiple fields → new MultiFieldPojoImplementation (returns a Record).

### DIFF
--- a/wayang-commons/wayang-basic/src/main/java/org/apache/wayang/basic/function/ProjectionDescriptor.java
+++ b/wayang-commons/wayang-basic/src/main/java/org/apache/wayang/basic/function/ProjectionDescriptor.java
@@ -38,8 +38,6 @@ public class ProjectionDescriptor<Input, Output> extends TransformationDescripto
     /**
      * Java implementation of a projection on POJOs via reflection.
      */
-    // TODO: Revise implementation to support multiple field projection, by names
-    // and indexes.
     private static class PojoImplementation<Input, Output>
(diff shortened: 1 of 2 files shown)


mspruc:
@zkaoudi I see you asked me for review, what's the intended usage here for POJOs? I think implementation in the PR is fine but I have some concerns about the type system, in the PR we wrap the output in `Record` but I assume that the output of the operator also wraps in `Record` would this be intended behaviour?

mspruc:
Hi, would it be possible for you to implement an integration test for you multi POJOs?

PR #711, opened 2026-03-05
Improve JdbcExecutor traversal by safely handling missing or multiple consumers

mohit-devlogs:
This change improves the robustness of JdbcExecutor by relaxing the assumption that the retrieval of consumers from the output channel is a singleton.
In the previous code, the function findJdbcExecutionOperatorTaskInStage() in the class WayangJdbcExecutor used the function WayangCollections.getSingle() for the retrieval of consumers from the output channel. The function throws an IllegalArgumentException if the list of consumers is empty or if there are more than one consumers in the list.

Changes:
- More robust code is now used in place of WayangCollections.getSingle()
- Return null if no consumer exists (i.e., it is the last in the pipeline)
- Throw a Wayang exception if there are more than one consumers

This prevents crashes like this:
IllegalArgumentException: [] is not a singleton

Testing:
- Compilation checked
- Tested with:
  mvn -pl wayang-platforms/wayang-generic-jdbc test

### DIFF
--- a/wayang-platforms/wayang-jdbc-template/src/main/java/org/apache/wayang/jdbc/execution/JdbcExecutor.java
+++ b/wayang-platforms/wayang-jdbc-template/src/main/java/org/apache/wayang/jdbc/execution/JdbcExecutor.java
@@ -1,3 +1,4 @@
+package org.apache.wayang.jdbc.execution;
 /*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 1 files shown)


novatechflow:
@mohit-devlogs - the PR conflicts with the current base, please rebase and follow up.

mohit-devlogs:
@novatechflow I recreated this change on top of the latest main because rebasing this branch resulted in extensive conflicts after upstream changes. The updated PR is #785. I'll close this PR so the review can continue there.

PR #713, opened 2026-03-06
Fix PythonApiTests Jenkins failure by using portable worker path configuration

### LINKED ISSUE #642: python api tests failing on jenkins
[PythonApiTests](https://github.com/apache/incubator-wayang/blob/main/wayang-api/wayang-api-python/src/test/java/org/apache/wayang/api/python/PythonApiTests.java) fails [jenkins](https://github.com/apache/incubator-wayang/blob/main/Jenkinsfile) build due to not providing a custom .properties with relevant paths.
The custom .properties will need to be formatted after [wayang.properties](https://github.com/apache/incubator-wayang/blob/b893e2cfd64c63b0a4bf2d283649908a661e1361/wayang-api/wayang-api-python/src/test/resources/wayang.properties#L4)

mohit-devlogs:
Fixes #642

The PythonApiTests test class was using a machine-specific absolute path for the Python worker script in the wayang.properties file, which was causing the Jenkins builds to fail in CI environments.

This PR changes the absolute path to a relative path, which is correct and portable, so that it can correctly resolve the path within the repository structure. The test class PythonApiTests was also using an unnecessary worker path override, which is removed in this PR so that it uses the test configuration provided in src/test/resources.

### DIFF
--- a/wayang-api/wayang-api-python/src/test/java/org/apache/wayang/api/python/PythonApiTests.java
+++ b/wayang-api/wayang-api-python/src/test/java/org/apache/wayang/api/python/PythonApiTests.java
@@ -31,17 +31,13 @@ class PythonApiTests {
     @Test
     void processCallerSocketTest() {
         try {
-        final Configuration configuration = new Configuration();
-        final String path = Paths.get("python/worker.py").toString();
-        
-        configuration.setProperty("wayang.api.python.worker", path);
+            final Configuration configuration = new Configuration();
+            final PythonProcessCaller processCaller = new PythonProcessCaller();
 
-        final PythonProcessCaller processCaller = new PythonProcessCaller();
+            assertTrue(processCaller.getSocket().isConnected());
+            assertTrue(processCaller.isReady());
 
-        assertTrue(processCaller.getSocket().isConnected());
-        assertTrue(processCaller.isReady());
-
-        processCaller.close();
+            processCaller.close();
         } catch (Exception e) {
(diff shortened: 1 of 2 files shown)


zkaoudi:
Thank you @mohit-devlogs. I think we had tried with the "../../" relative path solutions but it was still breaking. @mspruc can you confirm that?

mohit-devlogs:
> Thank you @mohit-devlogs. I think we had tried with the "../../" relative path solutions but it was still breaking. @mspruc can you confirm that?

Thanks for the clarification. I have used the relative path method for this PR. It means that the worker path will not be an absolute path from the CI environment; it will be a relative path from the project structure. It should also be noted if this method is different from what you were trying to implement earlier. I am also willing to change this if there are any known issues with this method. This method has already been tested, and the PythonApiTests are running successfully on Jenkins using this method. Earlier, they were not running using this method.

mspruc:
@zkaoudi I'm not sure i tried with `../../` in regards to jenkins? The reasons we had issues with configs originally was in regards to the CI workflows here, not on jenkins. 

@mohit-devlogs What steps have you taken to verify the worker works now that you've changed the test?

mohit-devlogs:
To verify this, the Python API test is run locally using the following command: 
mvn -pl wayang-api/wayang-api-python test

The build was completed successfully, and the test was able to locate the worker based on the relative path setup. The purpose of this modification is to avoid the use of paths in the environment. Instead, it is based on the relative paths.

mspruc (approved):

PR #721, opened 2026-03-17
[WAYANG-55] Refactor TableSource to support column projection.

tanyarai5446:
Currently, JdbcTableSource and other SQL-based operators cannot perform column projection because the parent TableSource class does not persist the columnNames array passed in the constructor.

Changes:
- Refactored TableSource to store columnNames as a protected field.
- Added getColumnNames() getter to TableSource.

Updated JdbcTableSource to utilize these names in createSqlClause for more efficient SQL generation (moving from SELECT * to SELECT col1, col2...).

Testing:
- Verified with mvn clean install -am -pl :wayang-jdbc-template (Build Success).


novatechflow:
@tanyarai5446 - the build failed, please check and rebase, then recommit.

PR #731, opened 2026-03-23
Add JavaParquetSink execution operator for the Java platform

gknz:
## Java Parquet Sink

Hey everyone! I built a Java platform execution operator for the existing `ParquetSink` logical operator, which previously only had a Spark implementation (`SparkParquetSink`). Now, the optimizer can choose between Java and Spark when writing Parquet files.

### What this PR adds

**JavaParquetSink** (`wayang-java/operators/`)
- Writes Wayang `Record`s to Parquet files using the `parquet-avro` library
- Follows the same pattern/logic as `SparkParquetSink`
- Infers Avro schema automatically by sampling up to 50 records
- Uses `RecordType` field names when available, falls back to `field0`, `field1`, etc.
- Uses Snappy compression
- Handles file overwrite mode

**ParquetSinkMapping** (`wayang-java/mapping/`)
- Connects the logical `ParquetSink` to `JavaParquetSink`
- Registered in Java platform `Mappings.java`

**Fluent API** (`DataQuantaBuilder.scala`)
- Added `writeParquet` method to `DataQuantaBuilder` so users can call it from the fluent API (from my understanding, this was missing — only the `DataQuanta.scala` layer existed)

**Unit Tests** (`JavaParquetSinkTest.java`)
- `testWriteStringRecords`: verifies basic write and read-back
- `testWriteMixedTypeRecords`: verifies type inference (Int, String, Double)
- `testWriteWithRecordType`: verifies column names from RecordType are preserved in the Parquet schema
- `testOverwriteExistingFile`: verifies overwrite mode replaces data

### Testing
All 4 unit tests pass. I also tested the operator in a different pipeline (not included here) that writes hourly aggregated data to Parquet and reads it back for further processing, and it was working as intended,

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,4 +35,4 @@ pom.xml.*
 # Scala Plugin for VSCode
 .metals
 .bloop/
-.metals/
+
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,5 @@ pom.xml.*
 
 # Scala Plugin for VSCode
 .metals
+.bloop/
+.metals/
--- a/wayang-api/wayang-api-scala-java/src/main/scala/org/apache/wayang/api/DataQuantaBuilder.scala
+++ b/wayang-api/wayang-api-scala-java/src/main/scala/org/apache/wayang/api/DataQuantaBuilder.scala
@@ -540,6 +540,36 @@ trait DataQuantaBuilder[+This <: DataQuantaBuilder[_, Out], Out] extends Logging
   }
 
   /**
+    * Feed the built [[DataQuanta]] into a [[org.apache.wayang.basic.operators.ParquetSink]]. This triggers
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -34,3 +34,5 @@ pom.xml.*
 
 # Scala Plugin for VSCode
 .metals
+.bloop/
+.metals/
```
zkaoudi: maybe we do not need the metals twice

zkaoudi (changes requested):

.gitignore
gknz: Hello @zkaoudi, I removed the duplicate.

zkaoudi (approved):

zkaoudi:
Thank you @gknz and congrats on your first contribution :)

PR #749, opened 2026-04-15
Simplify condition concatenation in SQL generation for JdbcExecutor

mohit-devlogs:
The PR makes handling filters while writing SQL queries in the JdbcExecutor easier.

Instead of having to consider adding separators while looping, `String.join` will be used to join filters, resulting in clearer code and reduced risk of errors in formatting.

It doesn’t have any impact on how the query functions but makes it more understandable.


zkaoudi (approved):

zkaoudi:
thank you @mohit-devlogs :)

PR #763, opened 2026-05-30
Add Short and BigDecimal type mappings to SqlTypeUtils

gknz:
## Summary
Adds proper SQL type support for `Short` and `java.math.BigDecimal` across the
table-sink write path. Previously neither had a JDBC type mapping, so the sink
created such columns as `VARCHAR(255)` (numbers stored as text, with no numeric
ordering/aggregation in the DB). This PR maps them to real numeric SQL types and
binds their values correctly on both the Java and Spark execution paths.

## Changes

**1. Type mapping (`SqlTypeUtils`)**
- `Short` / `short` → `SMALLINT`
- `BigDecimal` → `NUMERIC(38,18)`
- The PostgreSQL dialect inherits both from the default map.

**2. Value binding (the DDL type alone is not enough)**
Mapping the column type without binding the value would bind a *string* into a
numeric column. Each in-JVM sink is fixed:
- `JavaTableSink`: bind via `PreparedStatement.setShort(...)` / `setBigDecimal(...)`
  instead of the `setString` fallback.
- `SparkTableSink`: map `Short → DataTypes.ShortType` and
  `BigDecimal → DataTypes.createDecimalType(38, 18)`, so the DataFrame schema
  (and the JDBC write derived from it) produces `SMALLINT` / `NUMERIC(38,18)`.

## Why `NUMERIC(38,18)` rather than a bare `NUMERIC`
A bare `NUMERIC` is **not portable for decimals**. Per the SQL standard, omittting
the scale means scale 0, so most engines create an *integer* column and silently
truncate the fractional part: H2, MySQL (`DECIMAL(10,0)`), SQL Server
(`DECIMAL(18,0)`), Derby, DB2 all do this (e.g. `12.345 → 12`). PostgreSQL is the
lenient exception (bare `NUMERIC` is unbounded), which can hide the bug. An
explicit scale avoids the truncation on every database.

## Why specifically `38` and `18`
These are **not arbitrary**: `(38, 18)` is Spark's `DecimalType.SYSTEM_DEFAULT` —
the precision/scale Spark assigns to a decimal when none is otherwise known
(`38` is also the maximum precision Spark's `Decimal` supports, and a value widely
supported by Oracle/SQL Server). We deliberately reuse the **same** values on the
Java/DDL path (`SqlTypeUtils → NUMERIC(38,18)`):
- `38` = max total significant digits (generous headroom),
- `18` = digits after the decimal point (ample for monetary/rate data),
- and, crucially, this makes the **Java and Spark sinks emit identical column
  types** for a `BigDecimal`. In Wayang the optimizer decides which platform runs
  the sink, so the output schema must not depend on that choice — matching Spark's
  default keeps both paths consistent.

Trade-off: a fixed scale pads short values with trailing zeros
(`12.345 → 12.345000000000000000`). This is numerically identical and is exactly
what the Spark path already produced;=, in return decimals aree portable and lossless
across all databases.

## Tests
- `SqlTypeUtilsTest`: the full default type map (incl. `Short → SMALLINT`,
  `BigDecimal → NUMERIC(38,18)`) and PostgreSQL inheritance.
- `SparkTableSinkTest`:
  - direct mapping checks for `getSparkDataType(...)` — the whole Java-class →
    Spark `DataType` table, the `String` fallback, and explicitly
    `BigDecimal → DecimalType(precision=38, scale=18)`;
  - an end-to-end write of all supported types to in-memory H2, asserting the
    round-tripped values and the created column types (`SMALLINT`, `NUMERIC(38,18)`).
- `JavaTableSinkTest`: the analogous end-to-end test for the Java sink.

All tests run against in-memory H2 — no external database required.

## Notes
- The in-database JDBC sinks (`PostgresTableSinkOperator`, etc.) are unaffected :
  they compose `CREATE TABLE ... AS SELECT` and let the DB infer types, so they
  never bind a Java value or use `SqlTypeUtils`.
- `getSparkDataType` was widened from `private` to package-private only to make
  the mapping unit-testable.

### DIFF
--- a/wayang-platforms/wayang-java/src/main/java/org/apache/wayang/java/operators/JavaTableSink.java
+++ b/wayang-platforms/wayang-java/src/main/java/org/apache/wayang/java/operators/JavaTableSink.java
@@ -59,6 +59,10 @@ public class JavaTableSink<T> extends TableSink<T> implements JavaExecutionOpera
             ps.setDouble(index, (Double) value);
         } else if (value instanceof Float) {
             ps.setFloat(index, (Float) value);
+        } else if (value instanceof Short) {
+            ps.setShort(index, (Short) value);
+        } else if (value instanceof java.math.BigDecimal) {
+            ps.setBigDecimal(index, (java.math.BigDecimal) value);
         } else if (value instanceof Boolean) {
             ps.setBoolean(index, (Boolean) value);
         } else if (value instanceof java.sql.Date) {
--- a/wayang-commons/wayang-basic/src/main/java/org/apache/wayang/basic/util/SqlTypeUtils.java
+++ b/wayang-commons/wayang-basic/src/main/java/org/apache/wayang/basic/util/SqlTypeUtils.java
@@ -19,6 +19,7 @@
 package org.apache.wayang.basic.util;
 
 import org.apache.wayang.basic.data.Record;
+import java.math.BigDecimal;
 import java.sql.Date;
 import java.sql.Timestamp;
 import java.time.LocalDate;
@@ -42,10 +43,13 @@ public class SqlTypeUtils {
         defaultMap.put(int.class, "INT");
         defaultMap.put(Long.class, "BIGINT");
         defaultMap.put(long.class, "BIGINT");
+        defaultMap.put(Short.class, "SMALLINT");
+        defaultMap.put(short.class, "SMALLINT");
         defaultMap.put(Double.class, "DOUBLE");
         defaultMap.put(double.class, "DOUBLE");
         defaultMap.put(Float.class, "FLOAT");
         defaultMap.put(float.class, "FLOAT");
+        defaultMap.put(BigDecimal.class, "NUMERIC");
         defaultMap.put(Boolean.class, "BOOLEAN");
         defaultMap.put(boolean.class, "BOOLEAN");
         defaultMap.put(String.class, "VARCHAR(255)");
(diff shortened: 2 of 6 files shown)


wayang-platforms/wayang-java/src/main/java/org/apache/wayang/java/operators/JavaTableSink.java
```diff
@@ -59,6 +59,10 @@ private void setRecordValue(PreparedStatement ps, int index, Object value) throw
             ps.setDouble(index, (Double) value);
         } else if (value instanceof Float) {
             ps.setFloat(index, (Float) value);
+        } else if (value instanceof Short) {
+            ps.setShort(index, (Short) value);
+        } else if (value instanceof java.math.BigDecimal) {
```
mspruc, line 64: If you import `BigDecimal`, I don't think you have to specify `java.math.`, unless there's some other reason for this?

mspruc: Also, we are on Java 17, you can use pattern matching and do 
`if (value instanceof BigDecimal decimal) ps.setDecimal(index, decimal);`

mspruc:
Hi thanks for the contribution, if possible it would also be nice to have a unit test for this

zkaoudi:
@mspruc what unit test do you have in mind? This PR seems to simply add two more data types

mspruc:
@zkaoudi well something that tests the sinks ideally, e.g. for https://github.com/gknz/wayang/blob/efb35979111a6bb84192acd2ade8424bcd1ee463/wayang-platforms/wayang-spark/src/main/java/org/apache/wayang/spark/operators/SparkTableSink.java#L156 

we have two magic numbers, I'd like to make sure this conversion is stable.

gknz:
@mspruc hey! I've added unit tests for sqlTypeUtils, SparkTableSink, and JavaTableSink. Please feel free to read the updated descriptions for all the changes/additions I went with and why. If you have any comments, please feel free to let me know!

mspruc (approved):

mspruc:
@zkaoudi merge if you think it's ready

PR #772, opened 2026-06-28
Add Trino platform with engine-only integration tests

JunWang222:
## Summary

This PR adds the Wayang Trino platform implementation and engine-only integration tests.

## What is included

- Adds `wayang-trino` platform module.
- Adds Trino mappings/operators for:
  - TableSource
  - Filter
  - Projection
  - Join
  - GlobalReduce
  - ReduceBy
  - Sort
  - TableSink
- Adds end-to-end `TrinoOperatorsIT` coverage for each operator.
- Adds JavaPlanBuilder integration tests that simulate user-facing API usage.
- Keeps the integration tests engine-only by registering only `Trino.plugin()`.
- Uses self-contained test fixtures created by the test setup, without requiring external datasets.

## Validation

- Rebased on latest `apache/wayang:main`.
- Trino operator tests are designed to run against the local Trino setup and skip when Trino is unavailable.


zkaoudi:
Thank you for Jun. Can you do the following structural changes:
1) create a root folder platforms-setup-guides
2) move the "trino-setup" inside the above folder

zkaoudi:
Also, what does the the demo.sh demonstrate? Maybe we should also place it in another folder instead of having it in the root of the project.

zkaoudi:
sorry, I meant the demo-trino.sh

zkaoudi (approved):

PR #773, opened 2026-06-28
Add Presto platform with engine-only integration tests

JunWang222:
## Summary

This PR adds the Wayang Presto platform implementation and engine-only integration tests.

## What is included

- Adds `wayang-presto` platform module.
- Adds Presto mappings/operators for:
  - TableSource
  - Filter
  - Projection
  - Join
  - GlobalReduce
  - ReduceBy
  - Sort
  - TableSink
- Adds end-to-end `PrestoOperatorsIT` coverage for each operator.
- Adds JavaPlanBuilder integration tests that simulate user-facing API usage.
- Keeps the integration tests engine-only by registering only `Presto.plugin()`.
- Uses self-contained in-memory Presto fixtures created by the test setup, without requiring external datasets.

## Validation

- Rebased on latest `apache/wayang:main`.
- Presto operator tests are designed to run against the local Presto setup and skip when Presto is unavailable.

### DIFF
--- a/INTEGRATION-TESTS.md
+++ b/INTEGRATION-TESTS.md
@@ -1,4 +1,22 @@
-# Integration Tests — Presto
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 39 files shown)


mspruc:
Thank you for your contribution, looks good! left some comments.

zkaoudi:
Hi Jun, there is some test failing in the sql api. Could you take a look? I think it's the same in all three PRs

novatechflow (approved):
thank you @JunWang222

PR #780, opened 2026-07-05
Add research publications and citation section to README

zkaoudi:
Added a new section on research publications and citation for Apache Wayang so that researchers using our system can cite the right paper besides linking the repo.


kbeedkar (approved):
LGTM

PR #785, opened 2026-07-24
Improve JdbcExecutor traversal by safely handling missing or multiple consumers

mohit-devlogs:
This PR makes the JdbcExecutor more robust by ensuring that it handles situations where there are either zero consumers or multiple consumers of the execution task gracefully.

Change:

Returns null if the output channel has zero consumers rather than assuming that there is exactly one consumer.
Throws a helpful WayangException if there are multiple consumers of the execution task rather than making any unchecked assumptions.
Keeps the existing behavior intact for the typical case of having exactly one consumer.

This PR replaces my earlier PR for the same change. The branch has been recreated from the latest main after upstream refactoring made rebasing the previous branch impractical.


JunWang222 (commented):
Thanks for the update! Could you please add a small regression test for the multiple-consumer case?

mohit-devlogs:
@JunWang222 Thank you for the suggestion. I've added a regression test covering multiple consumer case, and JDBC module tests are passing.

PR #787, opened 2026-07-26
Add Trino cost profiling support and learned default parameters

JunWang222:
This PR adds Trino cost profiling support, including:

- A Trino profiling pilot for running operator cost profiling locally
- Shared profiling documentation that explains how users can rerun profiling on their own machine
- Trino-specific setup notes and command examples
- Default Trino cost parameters for users who want to try Wayang with Trino

The default parameters are intended as starting values. For accurate optimization on a specific machine or deployment, users should follow the profiling guide and regenerate parameters locally.


zkaoudi:
@JunWang222 some conflicts here too

wayang-tests-integration/src/test/java/org/apache/wayang/tests/FlinkIntegrationIT.java
```diff
@@ -314,12 +314,13 @@ void testReadAndTransformAndWriteWithIllegalConfiguration3() {
...
 
-        // Instantiate Wayang and activate the Spark backend.
+        // Instantiate Wayang and activate the Java/Flink backends.
         WayangContext wayangContext = makeContext(FLINK);
 
         // Have Wayang execute the plan.
         final Job job = wayangContext.createJob(null, wayangPlan);
-        // ILLEGAL: We blacklist the Spark platform, although we need it.
+        // ILLEGAL: We blacklist all platforms that can execute the plan.
+        job.getConfiguration().getPlatformProvider().addToBlacklist(Java.platform());
```
zkaoudi, line 323: Java is added here to the black list, but only flink is added in the context (line 318). Do we need to add Java to the black list too?

guides/cost-profiling.md
```diff
@@ -0,0 +1,430 @@
...
+  "p":0.9
+}
+```
+
+This can be read as:
+
+```text
+CPU load = alpha * number_of_rows + beta
+```
+
```
zkaoudi, line 89: we can add here that alpha=10 and beta=800000 in the example configuration above.

zkaoudi (changes requested):
please check some minor comments. Thank you @JunWang222 :)

wayang-tests-integration/src/test/java/org/apache/wayang/tests/FlinkIntegrationIT.java
JunWang222: The intention of this test is to register both Java and Flink, then blacklist both executable platforms so that only the made-up platform remains and the job fails as expected. I will make this explicit by changing the context setup from makeContext(FLINK) to makeContext(BOTH).

guides/cost-profiling.md
JunWang222: I will add the number in the example, thank you!

zkaoudi (dismissed):

zkaoudi:
@JunWang222 can you resolve conflicts here?

zkaoudi (approved):

PR #789, opened 2026-07-26
Add Presto cost profiling support and default parameters

JunWang222:
This PR adds Presto cost profiling support, including:

- A Presto profiling pilot for running operator cost profiling locally
- Shared profiling documentation that explains how users can rerun profiling on their own machine
- Presto-specific setup notes and command examples
- Default Presto cost parameters for users who want to try Wayang with Presto

The default parameters are intended as starting values. For accurate optimization on a specific machine or deployment, users should follow the profiling guide and regenerate parameters locally.


wayang-tests-integration/src/test/java/org/apache/wayang/tests/FlinkIntegrationIT.java
```diff
@@ -314,12 +314,13 @@ void testReadAndTransformAndWriteWithIllegalConfiguration3() {
         // Build a Wayang plan.
         final WayangPlan wayangPlan = WayangPlans.readTransformWrite(WayangPlans.FILE_SOME_LINES_TXT);
 
-        // Instantiate Wayang and activate the Spark backend.
+        // Instantiate Wayang and activate the Java/Flink backends.
         WayangContext wayangContext = makeContext(FLINK);
```
zkaoudi: I think here it has to be BOTH

zkaoudi (changes requested):
minor request, plus conflicts to be resolved

presto-setup/README.md
```diff
@@ -0,0 +1,249 @@
...
+1. Build the Wayang Presto platform and run the shared JDBC SQL-generation tests.
+2. Run the Wayang Presto operator tests against the live local PrestoDB.
+
+Run the commands below from the repository root. Java 17 and Docker with Docker
+Compose are required; Maven is provided by the repository wrapper.
+
+The Presto cost-profiling branch is named `feature/presto-cost-profiling`:
+
+```bash
+git checkout feature/presto-cost-profiling
```
zkaoudi: Same here. No need to checkout

zkaoudi (changes requested):

presto-setup/README.md
JunWang222: modified, thanks！

zkaoudi (approved):

PR #792, opened 2026-08-09
Implement JDBC Driver and Server for Apache Wayang

makarandhinge:
This PR introduces a JDBC interface for Apache Wayang as part of Google Summer of Code 2026.

It includes the JDBC driver, client-server protocol, JDBC server, SQL execution integration, result handling, metadata support, and resource management.

A local demo is also included for running the JDBC server and client.

Detailed architecture, design decisions, and usage are documented in the Apache Wayang website blog.


mspruc:
Hi thanks for your contribution, 

@zkaoudi demo data is included in this project, how should we handle larger integration tests / demo data, without making Wayang too large?

wayang-api/wayang-api-sql/src/main/java/org/apache/wayang/api/sql/context/SqlCatalogMetadata.java
mspruc: I'm not sure that I understand the need for this wrapper, if it just wraps a list of `SqlSchemaMetadata` immutably, can't we just use an immutable list?

wayang-jdbc/wayang-jdbc-client/src/main/java/org/apache/wayang/jdbc/driver/WayangConnection.java
```diff
@@ -0,0 +1,594 @@
...
+
+    private final String jdbcUrl;
+
+    private final String userName;
+
+    private final Properties clientInfo = new Properties();
+
+    private final Set<WayangStatement> statements = new LinkedHashSet<>();
+
+    private volatile boolean closed;
```
mspruc: why are these volatile if you handle synchronization within the accessor functions?

zkaoudi:
Good question. I'm not sure. I checked in Spark and they have a "resources" folder inside the root "examples" directory. Shall we do something similar?

novatechflow:
> Good question. I'm not sure. I checked in Spark and they have a "resources" folder inside the root "examples" directory. Shall we do something similar?

+1 for examples/

zkaoudi:
maybe we can agree on a policy that such datasets should be very small in size (eg., few tens of rows) instead of full-fledged datasets

PR #803, opened 2026-09-04
[SQL / Core] Preserve tuple schema and validate types between SQL and Java operators

### LINKED ISSUE #747: Missing schema/type validation between SQL and Java operators leads to runtime ClassCastException
While working on a pipeline like PostgresTableSource → Join → SqlToStream → Java map → collect, I faced a runtime ClassCastException. The issue is happening because after the SQL join, the data is actually in a tuple-like structure (like Tuple2<Record, Record>), but during the SqlToStream conversion, this structure is not preserved properly and everything is treated as a Record.

Because of this, when I try to use the data in the Java map step (for example accessing fields or casting types), it assumes a wrong structure and crashes at runtime (e.g., String cannot be cast to Integer). There is no validation or check between SQL and Java stages, so the mismatch is not caught early and only fails during execution.

Ideally, the correct schema/type should be preserved (or validated) when moving from SQL to Java, so that such issues don’t happen at runtime.

for example - 

CREATE TABLE table_a (
    id INT,
    name TEXT
);

CREATE TABLE table_b (
    id TEXT,   -- different type
    value TEXT
);

Image

AliRana30:
Closes #747

When executing a pipeline involving SQL TableSources, a SQL Join, and downstream Java operators (e.g., `PostgresTableSource -> Join -> SqlToStream -> Java map -> collect`), a runtime `ClassCastException` occurred. 

### Root Cause
1. In Wayang, the output of a SQL `JoinOperator` is structured as `Tuple2<Record, Record>`. However, `SqlToStreamOperator` and `GenericSqlToStreamOperator` hardcoded their output to flat `Record` objects, discarding the tuple structure. When downstream Java operators expected `Tuple2<Record, Record>`, they received raw `Record` objects and threw runtime `ClassCastException`.
2. Neither `Channel.addConsumer` nor `DefaultChannelConversion.convert` performed early schema/type validation between channels and consumers, so mismatches were not caught during plan compilation.

### Changes
- **Generic Operators**: Made `SqlToStreamOperator`, `SqlToRddOperator`, and `GenericSqlToStreamOperator` generic (`<Type>`) with support for emitting `Tuple2<Record, Record>` via `Tuple2ResultSetIterator` and `rowToTuple2`.
- **Dynamic Split Detection**: Implemented `resolveLeftColumnCount` and `getTableColumnCount` using database metadata and operator schemas to accurately partition joined rows into left and right `Record` instances.
- **Dynamic Channel Conversions**: Updated `ChannelConversions` across all SQL platforms (`postgres`, `sqlite3`, `trino`, `presto`, `bigquery`, `generic-jdbc`) to pass the channel producer's data type.
- **Early Type Validation**: Enhanced `Channel.addConsumer` and `DefaultChannelConversion.convert` to validate `consumerInput.getType().isSupertypeOf(producerSlot.getType())` or adapt types via `adaptType(...)`, throwing an early `IllegalArgumentException` on genuine mismatches.
- **Tests**: Added tests for SQL join to Stream `Tuple2` evaluation with Java mapping in `SqlToStreamOperatorTest`, and channel type validation tests in `ChannelTypeValidationTest`.

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## How Has This Been Tested?
- `SqlToStreamOperatorTest` (4 tests passed, including `testJoinWithHsqldbYieldsTuple2` and `testTypeAdaptationAndValidation`).
- `ChannelTypeValidationTest` (3 tests passed, verifying type mismatch rejection and compatible consumer acceptance).
- `JdbcJoinOperatorTest` and `GenericJdbcJoinOperatorTest` passed with 0 failures.
- Clean compilation across all SQL platform modules.

### DIFF
--- a/wayang-commons/wayang-core/src/main/java/org/apache/wayang/core/optimizer/channels/DefaultChannelConversion.java
+++ b/wayang-commons/wayang-core/src/main/java/org/apache/wayang/core/optimizer/channels/DefaultChannelConversion.java
@@ -72,6 +72,18 @@ public class DefaultChannelConversion extends ChannelConversion {
     public DefaultChannelConversion(
             ChannelDescriptor sourceChannelDescriptor,
             ChannelDescriptor targetChannelDescriptor,
+            BiFunction<Channel, Configuration, ExecutionOperator> executionOperatorFactory) {
+        this(
+                sourceChannelDescriptor,
+                targetChannelDescriptor,
+                executionOperatorFactory,
+                "via " + executionOperatorFactory.getClass().getSimpleName()
+        );
+    }
+
+    public DefaultChannelConversion(
+            ChannelDescriptor sourceChannelDescriptor,
+            ChannelDescriptor targetChannelDescriptor,
             BiFunction<Channel, Configuration, ExecutionOperator> executionOperatorFactory,
             String name) {
         super(sourceChannelDescriptor, targetChannelDescriptor);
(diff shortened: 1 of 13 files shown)


AliRana30:
Miss @zkaoudi needed your review on this as well. Thanks

PR #808, opened 2026-09-09
Add semantic filters with abstractions for semantic models as a platform

juripetersen:
This serves as an alternative to #797, whereas an abstraction called `SemanticOperator` was introduced, allowing to define future operators like `SemanticJoin` or other operators with this schema. New Models, like `Ollama` in our example can be added like Platforms, as they are essentially just execution platforms for `SemanticOperator`s.


conf/flink/default.properties
mspruc, line 1: move this to a different PR

pom.xml
mspruc, line 1: Does this need to be in this PR?

mspruc (approved):
Looks good

